PR #25, opened 2014-11-22
Adding file system details to the diagnostic output

### LINKED ISSUE #23: include filesystem type?
It would be great if the output could include the type of filesystem (e.g. ext2, nfs, whatever).

A long-term solution is to add this information (comes from java7 FileStore api) to ES stats, but as a start, maybe we could include the output from `df -k` + `mount` or similar?

pickypg:
Closes #23


jjfalling:

Issue #33, opened 2014-12-10
Add flag to disable ?pretty


jjfalling:
Large clusters can produce huge cluster states when using ?pretty which may not be desirable. A flag should be added to disable this.

PR #36, opened 2014-12-19
Removing pretty modifier from URLs to reduce download size.

pickypg:
The result of #33, but I want to leave it open so that it can be closed in the Java version.

### DIFF
--- a/bin/es-plugin.properties
+++ b/bin/es-plugin.properties
@@ -1,2 +1,2 @@
 description=Support diagnostics plugin for elasticsearch
-version=1.0.1
+version=1.1.0
--- a/bin/support-diagnostics.ps1
+++ b/bin/support-diagnostics.ps1
@@ -111,31 +111,31 @@ Write-Host 'Getting version'
 Invoke-WebRequest $esHost -OutFile $outputDir/version.json
 
 Write-Host "Getting _mapping"
-Invoke-WebRequest $esHost'/_mapping?pretty' -OutFile $outputDir/mapping.json
+Invoke-WebRequest $esHost'/_mapping' -OutFile $outputDir/mapping.json
 
 Write-Host 'Getting _settings'
-Invoke-WebRequest $esHost'/_settings?pretty' -OutFile $outputDir/settings.json
+Invoke-WebRequest $esHost'/_settings' -OutFile $outputDir/settings.json
 
 Write-Host 'Getting _cluster/settings'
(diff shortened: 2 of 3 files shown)


jjfalling:
Looks like the tool jsonpp (https://github.com/jmhodges/jsonpp) works
rather well at making the files readable. I've tried it on a 95mb json file
and it was quite fast.

The sized I posted before were the uncompressed files. Post compression the
archive w/o ?pretty ended up being 9.4M with normal gzip. But we are also
trying to get the size of the json files down, not just the whole archive.

rmuir:
> Looks like the tool jsonpp (https://github.com/jmhodges/jsonpp) works rather well at making the files readable. I've tried it on a 95mb json file and it was quite fast.

This has a lot of disadvantages though: I don't even have 'go' on any of my computers :)

> But we are also trying to get the size of the json files down, not just the whole archive.

why?

pickypg:
What do people think about adding `?format=yaml` to these instead of re-adding `?pretty`?

Having gone through a bunch of these, I tend to agree that the flat and pretty JSON leaves something to be desired even with the reformatting tools available (slows everything down).

Switching to YAML could provide a win in terms of readability and easy compression.

```bash
$ curl localhost:9200/_nodes/stats?format=yaml
```

Give node stats a shot to see what you think (among others).

```yaml
$ curl localhost:9200/?format=yaml
---
status: 200
name: "Kormok"
cluster_name: "elasticsearch"
version:
  number: "1.4.2"
  build_hash: "927caff6f05403e936c20bf4529f144f0c89fd8c"
  build_timestamp: "2014-12-16T14:11:12Z"
  build_snapshot: false
  lucene_version: "4.10.2"
tagline: "You Know, for Search"
```

clintongormley:
@pickypg that'd mean adding a YAML decoder to the JS stats tool as well.  Personally I'm not a fan of YAML as it has way too many gotchas.

@rmuir you have Perl on your computers, don't you? ;)  Installing JSON::XS gives you the `json_xs` command line tool which will prettify: https://metacpan.org/pod/distribution/JSON-XS/bin/json_xs

rmuir:
@clintongormley thanks for the hint. actually both my computers have json_pp (which seems to be based on that) already somehow. this works just fine.

jjfalling:
Too many folks do not like the unformatted json. Opened #43 and it should be re-added once I merge some branches.

Issue #4, opened 2014-09-18
Remove rmdir $outputdir line or need some additional checks to prevent accidental deletion of important directories


ppf2:
While this (https://github.com/elasticsearch/elasticsearch-support-diagnostics/blob/master/bin/support-diagnostics.sh#L80) is a convenience feature (for removing the outputdir if host is not reachable), it is quite dangerous esp if the script is run with sudo or root privileges.  Had a scenario in the field where -H is specified but not reachable and the user happened to have -o set to the directory where ES is installed, as a result, that directory is removed and the ES installation is gone ...

```
#ensure we can connect to the host, or exit as there is nothing more we can do
connectionTest=`curl -s -S -XGET $eshost 2>&1`
if [ $? -ne 0 ]
then
    echo "Error connecting to $eshost: $connectionTest"
    rmdir $outputdir
    exit 1
fi
```

jjfalling:
Should be fixed in the 2.x release. If not please reopen.

PR #48, opened 2015-02-12
Add dumping of iptables/ip6tables rules on node. Useful for connectivity troubleshooting

joshuar:
Quick hack to grab iptables rules.  Note this only grabs the filter tables. If customers are doing any nat'ing we will need them to run additional commands (iptables/ip6tables -S -t nat) but we probably don't need these for most issues.

Added a function to log script output which is tar'ed up with other diagnostics.

If user does not specify an output directory for diagnostics, use mktemp (part of coreutils) to create a unique one to use. Helps mitigate some of the issues outlined in Issue #4.

### DIFF
--- a/bin/support-diagnostics.sh
+++ b/bin/support-diagnostics.sh
@@ -11,6 +11,12 @@
 
 timestamp=$(date +"%Y%m%d-%H%M%S")
 
+log() {
+    local _msg=$*
+    echo "$_msg"
+    echo "$_msg" >> "$output_log"
+}
+
 # set defaults
 eshost="localhost:9200"
 targetNode='_local'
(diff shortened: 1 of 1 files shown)


jjfalling:
Few comments. 
1) Grabbing iptables (not the default fw across the supported distros) would require this plugin to be run as root, correct?
2) The log function is an interesting addition and I think would be useful.

joshuar:
Good point, I was running the diags from the command-line as root.  We may
want to avoid running as root?  In any case, if the iptables commands fail,
nothing happens, but can also remove them from the PR if needed.

markwalkom:
@joshuar You also gotta sign the CLA ;)

PR #49, opened 2015-02-13
Adds _alias output to the diagnostic dump

masaruh:
Sometimes it's useful especially when aliases have filter and/or routing.


jjfalling:
looks like cluster state does not contain this data. ++ to adding it.

jjfalling:
Actually, could you move this around where the settings are grabbed? I don't think we should need to get the aliases multiple times.

masaruh:
Ah, true. Fixed.

jjfalling:
looks good. also added to 2.0

PR #54, opened 2015-02-26
Changing hot_threads from 10 to 10000 to ensure that all threads are retrieved

### LINKED ISSUE #53: Return every thread (not just top 10 hot ones)?
When we pull diagnostics it's often useful to see not just the hot threads but any threads stuck waiting on a lock / IO operation as well, but because we only pull the top 10 "hot" ones now we won't (necessarily) see the stuck ones consuming 0% cpu.

I think we should show all threads?

pickypg:
Closes #53


mikemccand:
LGTM, thank you!

PR #58, opened 2015-04-28
Adding lower level config gathering

markwalkom:
Adding things like uname, free, mount, df, ifconfig and the like gives us direct view into the OS.

I know that knowing the output from `uname -a` would have made at least one issue quicker and easier to resolve.


joshuar:
This seems to share some similar edits to #25, wonder if they can be merged or one replaces the other?

jjfalling:

PR #64, opened 2015-05-12
Add a simple logging function that writes all output to stdout and a log file

joshuar:
This creates a simple log function that just `echo`'s all log messages to screen as well as logging to a file, which is then `tar`'ed up as part of the support diagnostics.


pickypg:
LGTM

jjfalling:

PR #65, opened 2015-05-15
Add curl --max-time

joshuar:
This PR and sets a default maximum execution time for a call to `curl` in the *nix script to 30s. This helps when, for example a call to a particular REST API end-point (like `/_segments`) takes a long time.  Ideally, getting most diagnostics back minus these calls which timeout is better than the end-user forcibly killing the script or Ctrl-C'ing when a command is taking too long.

### DIFF
--- a/bin/support-diagnostics.sh
+++ b/bin/support-diagnostics.sh
@@ -16,6 +16,7 @@ eshost="localhost:9200"
 targetNode='_local'
 repeat=1
 interval=60
+curlTimeout=30
 
 # pick up command line options
 while [ $# -gt 0 ]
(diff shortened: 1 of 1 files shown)


bin/support-diagnostics.sh
```diff
@@ -49,6 +51,7 @@ EOM
 	-a)  authType=$2;;
 	-c)  authCreds=$2;;
 	-p)  password=$2;;
+        -t)  curlTimeout=$2;;
```
pickypg: Extra leading space (not critical... but OCD).

pickypg:
LGTM

joshuar:
Will be deprecated with the upcoming 2.0 release.

pickypg:
Did you intentionally close this PR rather than merge it in?

joshuar:
Closed, because I thought we agreed to wait for the 2.0 (Java) branch, which incorporates these fixes anyway?

PR #66, opened 2015-05-18
Add -TimeoutSec option to Invoke-WebRequest calls

joshuar:
This commit adds a new option to the Windows script, `-t` allowing to
set a timeout (in seconds) for the REST API requests to Elasticsearch in
case these take a long time or hang.  This is done through the
`-TimeoutSec` option to the `Invoke-WebRequest` cmdlet.


joshuar:
Will be deprecated with the upcoming 2.0 release.

PR #127, opened 2016-09-26
Check permissions in diagnostics.sh

andersje:
diagnostics.sh needs read permission to /etc/elasticsearch and
/var/log/elasticsearch, or it will fail with an error partway through
the run.  This PR adds a permissions check to augment the note in the README.


andersje:
retest this please.

PR #128, opened 2016-09-26
Check permissions in diagnostics.sh

andersje:
diagnostics.sh needs read permission to /etc/elasticsearch and
/var/log/elasticsearch, or it will fail with an error partway through
the run.  This PR adds a pair of checks to the diagnostics.sh
script to validate that the two listed directories are readable.
It won't catch every permutation (e.g., if the directory is readable
and the files are not, it will fail), but it's a start.

### DIFF
--- a/scripts/diagnostics.sh
+++ b/scripts/diagnostics.sh
@@ -13,6 +13,17 @@ if [ ! -x "$JAVA" ]; then
     exit 1
 fi
 
+if [ ! -r "/etc/elasticsearch" ]; then
+    echo "Could not read config dir (/etc/elasticsearch).  Please verify permissions and rerun"
+    exit 1
+fi
+
+if [ ! -r "/var/log/elasticsearch" ]; then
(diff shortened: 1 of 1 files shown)


nemonster:
> The application can be run from any directory on the machine. It does not require installation to a specific location, and the only requirement is that the user has read access to the Elasticsearch artifacts, write access to the chosen output directory, and sufficient disk space for the generated archive.

I think I'm going to pass on this one...

The limitations are documented.  And if you review them further, you'll find that you can run this in remote mode where logs and configs are not collected.  As such, the diagnostics.sh is not the appropriate place to make this kind of check.  If you were even going to make this check, which I'm not so sure is necessary anyway.  The log collection piece will fail, but in most cases it appears to still collect the rest of the diagnostic output.  And will log the error, also as documented.

If you have a case where the permissions are insufficient and the entire utility fails, or the error isn't logged, please raise an issue and provide the steps to replicate and I'll take a look.

PR #142, opened 2017-01-27
-s is not a valid option

loekvangool:


nemonster:
Actually, it does work.  I just tried it. --ssl does work as well, and I've added that to the readme, but the change as submitted isn't valid.

PR #147, opened 2017-04-06
Cleanup shell script

frioux:
This makes the shell script slightly more efficient and more importantly allows you to run it from another directory

### DIFF
--- a/scripts/diagnostics.sh
+++ b/scripts/diagnostics.sh
@@ -17,5 +17,6 @@ fi
 
 [[ ${DIAG_JAVA_OPTS} == "" ]] && export DIAG_JAVA_OPTS="-Xms256m -Xmx2000m"
 
+diag_path="$(dirname $0)"
 echo "Using ${DIAG_JAVA_OPTS} ${DIAG_DEBUG_OPTS} for options."
-exec "$JAVA" $DIAG_JAVA_OPTS ${DIAG_DEBUG_OPTS} -cp .:./lib/*  com.elastic.support.diagnostics.DiagnosticApp "$@"
+exec "$JAVA" $DIAG_JAVA_OPTS ${DIAG_DEBUG_OPTS} -cp "$diag_path:$diag_path/lib/"*  com.elastic.support.diagnostics.DiagnosticApp "$@"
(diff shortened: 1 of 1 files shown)


frioux:
Ok I signed the CLA; how do we get it to update?

frioux:
Why did you close this instead of merging it?

--

nemonster:
This is a support tool. It's used commercially. It's public so you are free to look at it and verify exactly what we're doing. Or fork it and modify for your own use if you choose. But on the other hand with that in mind changes are generally made on the basis of a need by a support engineer, or to fix something that's obviously broken.

In terms of the exec change, there's never been an issue where saving that additional process is a concern. And because I interact with the system console quite a bit that means there's an extra round of testing I'd need to go through to make sure there weren't cases where this broke anything. I don't see a discussion in the submittal that would address that concern. You're leaving it to me to determine that.

In terms of the running from another directory I have a couple of issues. First, running only from the installation directory is by intent. Again, it's a tool we use for support. I want a narrow set of execution requirements that produce an "everybody does it this way so it usually works, if it doesn't there are a limited set of variables to examine, and you know where the archive is," result. It's not a feature set race to provide a competitive set of options. Changes are considered on the base of benefit/risk. 

In addition, I don't see a change for, or even a mention of the Windows bat file. Will that work the same? Does it now now need two sets of instructions, one for Win and one for the 'Nix environment? This doesn't appear to have been considered. And as such I have some concerns about whether the whole change would now break in certain cases that you just didn't consider when doing it. And again, there are issues of time that preclude me for going back to fully investigate something I'm not sure is warranted anyway.

If you'd like to fork it and modify for your own personal use, feel free. That's part of why it's public - it's fully transparent what we collect and how. And if someone offers a change/fix that either provides an immediate and recognizable benefit where you go, "Well that's a big enough deal that I don't mind allocating extra time," or it looks really, really low risk (like somebody adding code tags to the readme) then I'll probably merge. 

But this change just doesn't seem to fit those criteria. Sorry. Should have added this explanation when I closed. My apologies.

frioux:
Ok. I made the path change because I was actually running the tool during
an outage and couldn't run the tool unless I downloaded and extracted my
own copy. Oh well.

--

PR #150, opened 2017-05-01
Add Dockerfile to make usage easier without installing additional req…

calebcase:
…uirements.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:16.04
+
+RUN apt-get update && \
+    apt-get install -y \
+      curl \
+      default-jre \
+      unzip \
+    && \
+    apt-get clean
+
+WORKDIR /opt/elasticsearch-support-diagnostics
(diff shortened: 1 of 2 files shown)


calebcase:
CLA is signed, but it still shows here as not.

calebcase:
Nevermind... apparently you just have to comment and it will recheck ;o]

calebcase:
Would still be handy to have this.

nemonster:
Can you elaborate on why you even need this considering you can run it remotely? If you do that you can run it from a workstation.

calebcase:
@nemonster Sure! I run this from my own machine, but I don't want to install all those dependencies on my local machine. In particular, I don't otherwise have a for Java to be installed on this system.

IMHO, this makes it easier to provide reproducible environment to run the tool in that anyone with docker installed can use.

nemonster:
Still not sure I understand. You don't install any dependencies. You download a zip, and if you have a JRE from 1.7 on installed (which is usually a default on most workstations) that's it. There isn't a reproducible environment involved unless you're counting unzipping the archive.

calebcase:
I don't have a JRE installed so this helped me get it running without needing to install tools I wouldn't need except for when using this tool.

PR #151, opened 2017-05-09
Add flat_settings parameter to /_cluster/settings 

joshuar:
This output becomes a lot more readable to a human and potentially easier to search when `flat_settings` is specified.


jpcarey:
I just use the flatten json plugin for atom.io. You can flatten and unflatten no problem.

PR #155, opened 2017-06-09
add `bytes=b` for shards.json

jpcarey:
adding `bytes=b` provides a consistent size in the shards.json, making it easier to iterate for calculations rather than need to parse a string for `kb|mb|gb..`.


nemonster:
I've got a couple of other things to put in - will merge next week.

PR #162, opened 2017-07-11
Add jstack to diagnostics dump

jasontedor:
Having the output from jstack in diagnostics dumps would be incredibly helpful so this commit adds that.


jasontedor:
Thanks @nemonster!

PR #163, opened 2017-07-11
Throw helpful error message on unsupported major

jasontedor:
Today if the diagnostics tool is run against a major version that it does not support it throws a null pointer exception and does not really reveal what the problem is. This commit traps this situation and provides a helpful error message to the user.

Additionally, the logic for parsing the major version is broken as it will fail when the major version has two digits. This commit fixes this situation too.


jasontedor:
Thanks again @nemonster!

nemonster:
Should actually thank you for fixing something I should have addressed prior.

PR #164, opened 2017-07-12
Retrieve compressed logs by default

jasontedor:
Starting in Elasticsearch 6.0.0, when logs are rolled they will be compressed by default. This means that a support diagnostics dump should obtain these compressed logs too, otherwise a support diagnostics dump could be missing valuable logs for diagnosing an issue.

Relates elastic/elasticsearch#25660

### DIFF
--- a/src/main/java/com/elastic/support/diagnostics/InputParams.java
+++ b/src/main/java/com/elastic/support/diagnostics/InputParams.java
@@ -41,7 +41,7 @@ public class InputParams {
    private boolean isBzip = false;
 
    @Parameter(names = {"--archivedLogs"}, description = "Get archived logs in addition to current ones if present - No value required, only the option.")
-   private boolean archivedLogs=false;
+   private boolean archivedLogs=true;
 
    @Parameter(names = {"--scrub"}, description = "Set to true to use the scrub.yml dictionary to scrub logs and config files.  See README for more info.")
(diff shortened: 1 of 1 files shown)


nemonster:
I think I'm going to put this one to a support team vote before I do anything about it. That setting to not pull rolled logs when specified was put in by request. There's potentially a lot of output there and we were getting diags that were either too big to attach or took a while to upload/download. 

The output gets compressed, so is having the archived logs in compressed format going to buy you much other than possible speeding up the packaging?  Doesn't seem like there would be an appreciable size difference.

I can appreciate you want as much information as possible. But if it's a high severity incident you're playing beat the clock to some degree and that extra few minutes can make a difference both at the customer and the support level. I've already asked customers to run with the archived option when the current logs didn't suffice. 

So this is one I'll raise at the next group meeting.

I'm getting ready to do a few other things though.  Let me ask you this.  If it didn't pull everything but checked for the previous couple of versions or so as long as they didn't go over a certain size threshold would that be OK?

jpcarey:
I would vote to have the last three days by default. My reasoning is that I often get a diagnostic where rollover happened ~1h before diagnostic collection. Getting the previous day fixes that issue, and getting 2 days prior helps understand how long the issue might have been happening. That is my 2¢

PR #166, opened 2017-08-01
Added call to the role mapping API

jakommo:
In 5.5 we introduced the [role mapping API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-role-mapping.html) in addition to the role_mapping file. This PR adds a call to grab the results of this API.

### DIFF
--- a/src/main/resources/diags.yml
+++ b/src/main/resources/diags.yml
@@ -140,6 +140,7 @@ restQueries-5:
   watcher_stats: "_watcher/stats/_all"
   shield_users: "_xpack/security/user?pretty"
   shield_roles: "_xpack/security/role?pretty"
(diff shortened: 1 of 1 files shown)


jakommo:
@nemonster since this API is only available starting from 5.5, will this be an issue when running against pre 5.5 versions?

nemonster:
It will error out on everything below 5.5.  Although at this stage, about half the stuff that gets run errors out for one reason or another.  The API's have continually changed in minor versions and a number of the system calls require sudo to work so there's always going to be a certain amount of wreckage in the diags. Will merge this in for the next release - won't be that long, since there appears to be an issue with the jstack call that got added last release that I just repro'd and need to fix.

PR #185, opened 2017-11-21
scrub password for "key" config keys

immon:
Azure Respository Plugin [1] uses "key" config property to store
sensitive data.

[1] https://www.elastic.co/guide/en/elasticsearch/plugins/current/repository-azure-usage.html


nemonster:
I'm adding this key, but I've decided to do it in a fashion other than just adding another value to the code. Since it's probably other values will wind up requiring this, I've moved the values to look for to the diags.ylm file and use that instead of hard coded values.

PR #211, opened 2018-07-19
Connection message fix

nemonster:
See commits.

### DIFF
--- a/src/main/java/com/elastic/support/diagnostics/commands/VersionAndClusterNameCheckCmd.java
+++ b/src/main/java/com/elastic/support/diagnostics/commands/VersionAndClusterNameCheckCmd.java
@@ -4,6 +4,7 @@ import com.elastic.support.diagnostics.InputParams;
 import com.elastic.support.diagnostics.chain.DiagnosticContext;
 import com.elastic.support.util.JsonYamlUtils;
 import com.elastic.support.util.RestExec;
+import com.elastic.support.util.SystemProperties;
 import com.fasterxml.jackson.databind.JsonNode;
 
 import java.util.Map;
@@ -28,9 +29,9 @@ public class VersionAndClusterNameCheckCmd extends AbstractDiagnosticCmd {
          context.setClusterName(clusterName);
          String versionNumber = root.path("version").path("number").asText();
          context.setVersion(versionNumber);
-
       } catch (Exception e) {
-         logger.info("Error retrieving Elasticsearch version  - unable to continue..  Please make sure the proper connection parameters were specified", e.getMessage());
+         logger.info(e.getMessage());
+         logger.info("Error retrieving Elasticsearch version  - unable to continue..  Please make sure the proper connection parameters were specified.");
          rc = false;
       }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/elastic/support/diagnostics/commands/VersionAndClusterNameCheckCmd.java
```diff
@@ -28,9 +29,9 @@ public boolean execute(DiagnosticContext context) {
          context.setClusterName(clusterName);
          String versionNumber = root.path("version").path("number").asText();
          context.setVersion(versionNumber);
-
       } catch (Exception e) {
-         logger.info("Error retrieving Elasticsearch version  - unable to continue..  Please make sure the proper connection parameters were specified", e.getMessage());
+         logger.info(e.getMessage());
+         logger.info("Error retrieving Elasticsearch version  - unable to continue..  Please make sure the proper connection parameters were specified.");
```
pickypg: Can you not do

```java
logger.info("message", e);
```

Or perhaps,

```java
logger.info(e, "message");
```

One variant should be valid and print the stacktrace with the exception messages.

src/main/java/com/elastic/support/util/RestExec.java
```diff
@@ -91,10 +91,10 @@ protected HttpResponse exec(String url) {
          }
       } catch (HttpHostConnectException e) {
          logger.log(SystemProperties.DIAG, "Host connection error.", e);
-         throw new RuntimeException("Error connecting to host " + url, e);
+         throw new RuntimeException("Host connection error: " + e.getMessage() );
```
pickypg: This drops the URL (unless it's always printed by the exception?) and also eliminates any caused by exceptions printable in the logs. I assume that's intentional?

nemonster: See above. The URL is actually in the message for this type of error. I've already logged the stack trace in the internal log.

src/main/java/com/elastic/support/diagnostics/commands/VersionAndClusterNameCheckCmd.java
nemonster: Actually no. In this case there are two separate logging appenders - one for the user that goes to the console and one that's being written to the diagnostic archive and sent up. The diag appearer is using a custom level to separate output that only comes back to us. In the past, showing customers that actual stack traces has ranged from having no impact to actually creating more issues. So I don't show them any of it, and at this stage create an archive with a manifest, any other artifacts and the log with the actual stack traces. 

There's more granular info if they actually get to running something. This is to address the case where they've  failed because they tried to use http instead of https. So in this case showing a user who's not aware they they are running secure sockets or need to specify a login and  password is unlikely to help much. But it does provide a bit more info in terms of the actual error for the support engineer.

pickypg: Ah, then in that case, then why print two separate messages?

nemonster: Because that's the user output and I wanted it on two separate lines. It looks better to me than appending the system line separator. Just my pref.

pickypg: Hard to argue with preferences.

Issue #177, opened 2017-10-05
Return shard level stats in indices_stats


PhaedrusTheGreek:
Would be useful for diagnosing hot spotting

```
indices_stats: "_stats?level=shards&pretty&human"
```

nemonster:
Do you want that as a separate output?

PhaedrusTheGreek:
Seems that built in would make sense - I don't see it breaking anything as from what I can tell it adds a new key for each index called `shards`

ppf2:
This file can be huge if there are tons of shards in the cluster (and using pretty).  We may also want to check with dev to see if it's expensive to ask for shard level metrics for all indices.  If so, users will have to be careful esp. when running on an interval.

nemonster:
Yes, I may make it a separate option where you specify it via a --type. So you'd have to explicitly call for it if you wanted that kind of output. Run at your own risk.

inqueue:
+1 I am finding more and more where shard level stats would be helpful to have; see the primary and replica divergence issue. I'd be fine to call it without `?pretty` as it can be prettified with `jq` or any text editor with a pretty JSON plugin. Further the way I understand it, Cat APIs are less efficient anyway as it calls shard level stats internally.

jpcarey:
@ppf2 json size should be the only perf concern. I was just chatting with @bleskes and @jasontedor, and they collect all this data for the call no matter what - it just a matter of what is output in the final json.

nemonster:
Incorporated into the upcoming 6.4

jpcarey:
I think we forgot to add it to all the versions. Currently, I only see `level=shards` for `standard-1`.

PhaedrusTheGreek:
I'm also not seeing this in  `"diagVersion" : "6.4.4"`

PR #222, opened 2018-08-31
Fix for ES shard level stats, and add request more threads for logstash

jpcarey:
Fixes https://github.com/elastic/support-diagnostics/issues/177

Shard level stats was supposed to be enabled for all ES version, but was only set `standard-1`. This adds it to `standard-2`, `standard-5`, and `standard-6`

Added `threads=10000` to logstash_nodes_hot_threads request. The default is 3 threads, which is pretty useless. To note, I checked with logstash team to get an ok for this:

> Jared Carey [12:03 PM]
would there be any negative impact on issuing `_node/hot_threads?human=true&threads=10000`? The support diagnostic utility currently only collects the default (3 threads). For elasticsearch, we collect with `threads=10000`. I wanted to check before I submit a PR to change this.

> Jordan Sissel [12:15 PM]
@\jared that seems ok especially given you already do it for Es.


jpcarey:
@nemonster just pinging to see if we can get this merged.

PR #223, opened 2018-09-10
Remove leading slashes

joegallo:
Most of the urls lack leading slashes, tidying things up so that's the same everywhere.


nemonster:
Removing the forward slashes from the url passed to the HttpClient turns out to break submissions when the diagnostic is executed against a cloud cluster. We have a customer who is using the remote option to get diags from their cloud deployment. In addition to that, it appears the the natural tendency of those who are submitting requests to add additional calls is to add the leading slash. Therefore, since from what I can ascertain this file is used as a template rather than directly I will be going back to forward slashes, although I will be checking to make sure they are added consistently.

Issue #229, opened 2018-11-07
Argument --heapdump options doesn't work


Leaf-Lin:
I am trying to get heapdump via support-diagnostics as described here: https://github.com/elastic/support-diagnostics/blob/270428ff59805c75396be682b92dd1193cd8a68a/README.md#L136

But get error message:
```
INFO  Error during diagnostic initialization: Unknown option: --heapdump
```

I know I can get heapdump via the `jmap` command instead, just wondering if this option in latest support-diagnostics readme should be hidden if it’s not working? or maybe I have misunderstand how to use this?

PR #234, opened 2019-01-02
Update README.md

geekpete:
- Inline code quotes for parameter names and values for readability
- Removed reference to Heap dump functionality until https://github.com/elastic/support-diagnostics/issues/229 is resolved.


geekpete:
The difference in readability is almost invisible, maybe bold would be a better idea for the parameters instead of inline code blocking them.

PR #235, opened 2019-01-07
Add watcher_stack back to commercial versions

hub-cap:
watcher_stack, the emit_stacktraces=true API call for watcher was not
added to newer versions of the diags tool. When diagnosing a cluster
recently, we wanted to look at these. This will add it back to all major
versions in the event we need to look at it next time.

### DIFF
--- a/src/main/resources/diags.yml
+++ b/src/main/resources/diags.yml
@@ -120,6 +120,7 @@ standard-2:
 commercial-2:
   licenses: "_license?pretty"
   watcher_stats: "_watcher/stats/_all?pretty"
+  watcher_stack: "_watcher/stats?emit_stacktraces=true"
   security_users: "_shield/user?pretty"
   security_roles: "_shield/role?pretty"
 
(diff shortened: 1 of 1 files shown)


hub-cap:
ping @spinscale 	since i cannot add reviewers

hub-cap:
just an fyi, we would also like to add it to cloud, should i just have them add it manually? They typically sync the same file here. I could ask them to diverge from what you have for now since you are refactoring.

What is the ETA of the refactor?

nemonster:
>>just an fyi, we would also like to add it to cloud, should i just have them add it manually? They typically sync the same file here. I could ask them to diverge from what you have for now since you are refactoring.

I don't know how to answer that since I have no idea how it's used by Cloud. If they have a need for a particular format I think it might be appropriate to describe the usage, as well as articulating why a change in that context.

> What is the ETA of the refactor?
Later this week, more than likely. Early next week latest.

hub-cap:
Cloud simply copies the file out of your repo and syncs it up manually to their repo. I could ask them to diverge from your code here, especially given that you will have it changed soon. Also, please add that call in to whatever you end up redoing. <3 cc @jakelandis 	so he is aware.

nemonster:
Closing since refactored configuration is will be merged shortly.

hub-cap:
<3 for https://github.com/elastic/support-diagnostics/blob/master/src/main/resources/diags.yml#L79

PR #239, opened 2019-01-10
Refactor diag config

nemonster:
1. Refactor and update the REST calls. Because they change now in minor releases rather than just at major, modified the process to take a subset of common calls and then build a full set appropriate for that version. This also allows the diag to be run against a new version not explicitly configured yet. In that case it will use the latest available set of commands. 

2. When calls return with a non-200 code, such as a 400 write the entire response output to the file as well as logging the occurrence. Requested by tech leads.

3. Minor changes and clarifications to README.


src/main/java/com/elastic/support/util/RestExec.java
```diff
@@ -93,20 +93,13 @@ else if(status == 403){
       return responseString;
    }
 
-   public boolean execConfiguredQuery(String url, String query, String destination) {
+   public void execConfiguredQuery(String url, String query, String destination) {
 
       HttpResponse response = null;
       boolean ret = true;
```
pickypg: Weird formatting in this function now that you removed the `try` wrappers. And this variable is no longer needed / used.

src/main/java/com/elastic/support/util/RestExec.java
```diff
@@ -140,36 +132,23 @@ protected String getResponseString(HttpResponse response) {
...
 
-   protected boolean streamResponseToFile(HttpResponse response, String destination) {
+   protected void streamResponseToFile(HttpResponse response, String destination) {
 
       try {
          int status = checkResponseCode(response);
          org.apache.http.HttpEntity entity = response.getEntity();
          InputStream responseStream = entity.getContent();
-
+         FileOutputStream fos = new FileOutputStream(destination);
```
pickypg: You should close the output stream after opening it (using Java 7's [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)):

```diff
+try (FileOutputStream fos = new FileOutputStream(destination)) {
+    IOUtils.copy(responseStream, fos);
+}
-FileOutputStream fos = new FileOutputStream(destination);
-IOUtils.copy(responseStream, fos);
```

src/main/java/com/elastic/support/util/RestExec.java
```diff
@@ -93,20 +93,13 @@ else if(status == 403){
       return responseString;
    }
 
-   public boolean execConfiguredQuery(String url, String query, String destination) {
+   public void execConfiguredQuery(String url, String query, String destination) {
 
       HttpResponse response = null;
       boolean ret = true;
-      try {
          response = exec(url);
```
pickypg: If `exec` throws an exception, then it will not write the response to disk. Should it for when `exec` rethrows something other than `HttpHostConnectException`?

It's unclear to me what would qualify as a "Query Execution Error" since it doesn't throw for HTTP 400, so I think we're safe and should perhaps just retain the `try/catch` with it silently ignoring exception as it already does?

pickypg: > Throwable would be even less discriminating I guess.

True, but you're not supposed to catch `Throwable` in most cases.

pickypg: I marked it as LGTM, but afterward it occurred to me: doesn't this code now [re]throw a `RuntimeException` where before it didn't throw exceptions?

nemonster: I actually went back took another look to make sure I knew what I was getting into because I had ripped out so much. On one hand, the original idea for this particular part was that a HostConnectionException or the possible Runtime Exception would be so egregious you'd want to just rethrow it and get out of the diag query process altogether. And that would get rethrown again further down the line shutting down the entire process.

But OTOH, thinking about it, it could be intermittent. A sporadic issue with a host or proxy. In which case the next one might work. So I'm going to do another pass and remove these while taking another step through what goes on.

Actually the getResponseString method also has this issue. You could have an issue getting a particular response but given the rethrow it would shout everything down. So I'm probably going to just log everything and try the next one. I can think of at least one person who wants everything to fail as soon as anything fails who isn't going to like that. But if there are issues the additional operations probably have diagnostic value as well when they send the failed archive in.

PR #249, opened 2019-01-29
Include ccr info, stats and auto follow pattern endpoints.

martijnvg:
CCR is going to be GA from 6.7, so it is time to include some of its endpoints in the diagnostic tool.

### DIFF
--- a/src/main/resources/diags.yml
+++ b/src/main/resources/diags.yml
@@ -120,6 +120,10 @@ rest-calls:
       minor-5:
         rollup_index_caps: "*/_xpack/rollup/data"
         security_priv: "_xpack/security/privilege?pretty"
+        ccr_stats: "_ccr/stats?pretty"
+        ccr_autofollow_patterns: "/_ccr/auto_follow?pretty"
+      minor-7:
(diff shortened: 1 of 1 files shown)


martijnvg:
I tested this by building the binary against a local cluster with ccr.

I also noticed that the `TestRestCalls` fails in the IDE, but not when running `mvn verify`:

```
INFO] --- maven-surefire-plugin:2.12.4:test (default-test) @ support-diagnostics ---
[INFO] Surefire report directory: /home/mvg/dev/projects/support-diagnostics/target/surefire-reports

-------------------------------------------------------
 T E S T S
-------------------------------------------------------

Results :

Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
```

Looks like this test never gets executed. Is this expected?

nemonster:
On that note, I won't be merging this for a few more days unless there's some reason it needs to go in sooner.

martijnvg:
> On that note, I won't be merging this for a few more days unless there's some reason it needs to go in sooner.

:+1: There is no rush and this can wait a few days.

PR #250, opened 2019-01-30
Sort _cat APIs by index

pickypg:
This adds the `s=index` parameter for 5.2+ clusters for these APIs:

```
_cat/indices
_cat/segments
_cat/shards
```

Sorting by index helps to make the separate files more consistent and easier to read by humans.


pickypg:
Thanks!

PR #257, opened 2019-03-31
Include shard level stats in indices stats

dnhatn:
This change adds the shard level stats such as commit stats, seq_no stats to the indices stats. These shard level stats are valuable especially in 6.0+.


nemonster:
Actually, before I do anything with this... 

How many versions back will this work?  IOW, does it start with 6.0? Further back? If so, how far?

With the newer configuration it does make a difference.

dnhatn:
@nemonster Thanks for reviewing. This change should work with any ES version because the shard level parameter is provided at the [beginning](https://github.com/elastic/elasticsearch/commit/80062fbe10b574cd1a9723caf197254a5c54e752#diff-9b34c2329f6af838fe501ee18cf61f34R177) of the indices stats API. I am fine if you make this change for 6.0+ only.

Issue #267, opened 2019-04-11
[7.0.x] diagnostics.sh RuntimeException: DiagnosticService runtime error 


inqueue:
Support diagnostics against 7.0.0 GA is resulting in `java.lang.RuntimeException`. Attempted with support diagnostics 7.0.2 and 7.0.0. 

```
cat /etc/redhat-release
CentOS Linux release 7.6.1810 (Core)

java -version
openjdk version "1.8.0_201"
OpenJDK Runtime Environment (build 1.8.0_201-b09)
OpenJDK 64-Bit Server VM (build 25.201-b09, mixed mode)
```

```
sudo ./diagnostics.sh --host localhost -u elastic -p --ssl --noVerify
Using /bin/java as Java Runtime
Using -Xms256m -Xmx2000m  for options.
15:59:07 INFO  Processing diagnosticInputs...
Password:
15:59:12 INFO  Creating temp directory: /home/user/support-diagnostics-7.0.2/diagnostics
15:59:12 INFO  Created temp directory: /home/user/support-diagnostics-7.0.2/diagnostics
15:59:12 INFO  Configuring log file.
15:59:12 INFO  Writing diagnostic manifest.
15:59:12 INFO  Getting Elasticsearch Version.
15:59:13 INFO  Log close complete.
Exception in thread "main" java.lang.RuntimeException: DiagnosticService runtime error
	at com.elastic.support.diagnostics.chain.DiagnosticChainExec.runDiagnostic(DiagnosticChainExec.java:22)
	at com.elastic.support.diagnostics.DiagnosticService.exec(DiagnosticService.java:57)
	at com.elastic.support.diagnostics.DiagnosticApp.main(DiagnosticApp.java:31)
```

Help also results in error (this command works with v.6.4.4):
```
./diagnostics.sh --help
Using /bin/java as Java Runtime
Using -Xms256m -Xmx2000m  for options.
16:05:57 INFO  Processing diagnosticInputs...
Exception in thread "main" java.lang.NullPointerException
	at com.elastic.support.config.DiagnosticInputs.validate(DiagnosticInputs.java:248)
	at com.elastic.support.diagnostics.DiagnosticApp.main(DiagnosticApp.java:21)
```

jguay:
--help also fails for me but I can generate a bundle from diag 7.0.2 against 7.0.0 cluster from the elasticsearch container or from my MacOS directly (elasticsearch running with self-signed certificate - same command as above)
JVM on my mac : 
```
[...]

PR #268, opened 2019-04-11
Fix Host Name Verification Bypass

nemonster:
- variable was being set with diagnostic version check bypass


nemonster:
Fixes #267

PR #273, opened 2019-04-26
Wrong package path to ScrubApp

lastnico:
Fix the packages in SH & BAT scripts

### DIFF
--- a/scripts/scrub.bat
+++ b/scripts/scrub.bat
@@ -14,6 +14,6 @@ if not defined DIAG_JAVA_OPTIONS (
   set DIAG_JAVA_OPTIONS=-Xmx2000m
 )
 
-"%JAVA_EXEC%" %DIAG_JAVA_OPTIONS% -cp .\;.\lib\* com.elastic.support.diagnostics.ScrubApp %*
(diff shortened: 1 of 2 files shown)


elasticcla:
Hi @lastnico, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/support-diagnostics/pull/273.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

lastnico:
I've just signed, afterwards, the Contributor Agreement, so if you can relaunch the checks on this pull request, that would be cool.

lastnico:
Ah, damn it, i've used the default proposed author on my IDE, which is not similar to my Github account. I guess I'll need to redo all to change it (it's not a proper email address).

nemonster:
I'm going to go ahead and approve this despite the CLA check since I could have just gone back and redone it myself and it's not anything inherently risky or compromising.

nemonster:
@lastnico When you get your CLA issue fixed, keep in mind that we'd like to see an issue that corresponds the pull request as well so we can track it better. Thanks for the patch!

pickypg:
@lastnico You could also rewrite your commits and do a force push with the proper author.

PR #276, opened 2019-05-10
Add compatibility 7.x in README

renshuki:
Currently 7.x isn't specified in the docs but I didn't find any particular issue running the support-diagnostics on that version. Is there any reason to not add 7.x to the README?

Thanks 🙏


nemonster:
Looked over this, and actually wanted to go beyond adding a version number since now the diagnostic works with any version of ES, and will use the latest release if it can't match the version number it retrieves from the cluster.

PR #286, opened 2019-06-17
Add check for sufficient privileges to run diag.

nemonster:
New command to check user for admin or superuser role, depending on version.
Version class added to handle major/minor/patch comparisons.
More explicit message and earlier exit if ES version could not be obtained.

### DIFF
--- a/src/main/java/com/elastic/support/diagnostics/commands/UserRoleCheckCmd.java
+++ b/src/main/java/com/elastic/support/diagnostics/commands/UserRoleCheckCmd.java
@@ -23,62 +23,44 @@ public class UserRoleCheckCmd implements Command {
 
         RestClient restClient = context.getEsRestClient();
         String user = context.getDiagnosticInputs().getUser();
+        boolean hasAuthorization = false;
 
         Version version = context.getVersion();
         String url = null;
         int major = version.getMajor();
         int minor = version.getMinor();
 
-        if(major == 2){
+        if (major <= 2) {
             url = "/_shield/user/" + user;
-        }
-        else if (major <= 6){
+        } else if (major <= 6) {
             url = "/_xpack/security/user/" + user;
-        }
-        else{
+        } else {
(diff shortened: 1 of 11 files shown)


src/main/java/com/elastic/support/diagnostics/commands/UserRoleCheckCmd.java
```diff
@@ -0,0 +1,85 @@
...
+            String userJsonString = result.toString();
+            JsonNode userNode = JsonYamlUtils.createJsonNodeFromString(userJsonString);
+            JsonNode rolesNode = userNode.path("roles");
+            List<String> roles = new ArrayList<>();
+            if(rolesNode.isArray()){
+                ObjectMapper mapper = new ObjectMapper();
+                roles = mapper.convertValue(rolesNode, List.class);
+            }
+
+            if(major < 6){
```
pickypg: This could be simplified to:

```java
// prior to the opening of X-Pack the "superuser" user role necessary was called "admin"
if (major < 6 || (major == 6 && minor < 3)) {
    hasAuthorization = roles.contains("admin");
} else {
    hasAuthorization = roles.contains("superuser");
}
```

src/main/java/com/elastic/support/diagnostics/commands/UserRoleCheckCmd.java
```diff
@@ -0,0 +1,85 @@
...
+
+        Version version = context.getVersion();
+        String url = null;
+        int major = version.getMajor();
+        int minor = version.getMinor();
+
+        if(major == 2){
+            url = "/_shield/user/" + user;
+        }
+        else if (major <= 6){
```
pickypg: This will be used for version `1`, which is never going to return anything useful for the given request since X-Pack and its APIs wasn't a thing until 5.0. Perhaps we should just skip this for `1.x` and log a warning saying it's unverified (and perhaps a tip to upgrade)?

PR #297, opened 2019-08-06
Id auth errors

### LINKED ISSUE #293: [7.0.8] RuntimeException with bad password
[1,0,0,1]

Dup of https://github.com/elastic/support-diagnostics/issues/285, but I do not have permission to reopen issues in this repo.

An authentication failure (bad credentials) to the cluster results in an uninformative `DiagnosticService runtime error`:

```
~/tmp/support-diagnostics-7.0.8
./diagnostics.sh -u elastic -p --ssl --host cluster-host --port 9200
Using /usr/bin/java as Java Runtime
16:38:52 INFO  Processing diagnosticInputs...
Password:
16:38:55 INFO  Creating temp directory: /tmp/support-diagnostics-7.0.8/diagnostics
16:38:55 INFO  Created temp directory: /tmp/support-diagnostics-7.0.8/diagnostics
16:38:55 INFO  Configuring log file.
16:38:55 INFO  Writing diagnostic manifest.
16:38:55 INFO  Checking for diagnostic version updates.
16:38:55 INFO  Getting Elasticsearch Version.
16:38:56 INFO  Log close complete.
Exception in thread "main" java.lang.RuntimeException: DiagnosticService runtime error
	at com.elastic.support.diagnostics.chain.DiagnosticChainExec.runDiagnostic(DiagnosticChainExec.java:22)
	at com.elastic.support.diagnostics.DiagnosticService.exec(DiagnosticService.java:59)
	at com.elastic.support.diagnostics.DiagnosticApp.main(DiagnosticApp.java:31)
```

Could the diagnostic return an error more descriptive for the cause?

nemonster:
Closes #293 
* Reworked error handling, moving some of the logging and checking down to the result level from the BaseQuery and changing the command execution to break out of the chain execution on a fatal error, such as a bad version check call. Add specific error codes to display in order to flag a bad login at the command line.

### DIFF
--- a/src/main/java/com/elastic/support/diagnostics/DiagnosticService.java
+++ b/src/main/java/com/elastic/support/diagnostics/DiagnosticService.java
@@ -27,6 +27,7 @@ public class DiagnosticService extends BaseService {
         // Create two clients, one generic for Github or any other external site and one customized for this ES cluster
         RestClient genericClient = createGenericClient(diagConfig, diagnosticInputs);
         RestClient esRestClient = createEsRestClient(diagConfig, diagnosticInputs);
+        DiagnosticContext ctx = new DiagnosticContext();
 
         try {
 
@@ -48,8 +49,6 @@ public class DiagnosticService extends BaseService {
             List<String> chain = chains.get(diagnosticInputs.getDiagType());
 
             DiagnosticChainExec dc = new DiagnosticChainExec();
-            DiagnosticContext ctx = new DiagnosticContext();
-
             ctx.setEsRestClient(esRestClient);
             ctx.setGenericClient(genericClient);
             ctx.setDiagsConfig(diagConfig);
@@ -58,9 +57,16 @@ public class DiagnosticService extends BaseService {
 
             dc.runDiagnostic(ctx, chain);
 
+            if(ctx.isBypassSystemCalls()){
+                logger.warn("Could not locate local node - bypassing log collection and system calls.");
+            }
+            else if (ctx.isDocker()){
+                logger.warn("Identified Docker installations - bypassing log collection and system calls.");
+            }
+
             String user = ctx.getDiagnosticInputs().getUser();
 
-            if(StringUtils.isNotEmpty(user) && !ctx.isAuthorized()){
+            if (StringUtils.isNotEmpty(user) && !ctx.isAuthorized()) {
 
                 String border = SystemUtils.buildStringFromChar(60, '*');
                 logger.warn(SystemProperties.lineSeparator);
@@ -75,16 +81,17 @@ public class DiagnosticService extends BaseService {
                 logger.warn(border);
 
             }
-
-            closeLogs();
-            createArchive(ctx.getTempDir());
-            SystemUtils.nukeDirectory(tempDir);
+        } catch (DiagnosticException de) {
+            logger.info("Fatal error in diagnostic - could not continue. See diagnostics.log in archive or temp directory for more details.");
+            logger.log(SystemProperties.DIAG, "Fatal error in commands", de);
 
         } catch (IOException e) {
             logger.error("Access issue with temp directory", e);
             throw new RuntimeException("Issue with creating temp directory - see logs for details.");
         } finally {
             closeLogs();
+            createArchive(ctx.getTempDir());
+            SystemUtils.nukeDirectory(ctx.getTempDir());
             esRestClient.close();
             genericClient.close();
         }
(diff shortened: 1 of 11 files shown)


src/main/java/com/elastic/support/diagnostics/commands/VersionCheckCmd.java
```diff
@@ -31,14 +33,19 @@ public void execute(DiagnosticContext context) {
         try {
             DiagnosticInputs diagnosticInputs = context.getDiagnosticInputs();
             RestClient restClient = context.getEsRestClient();
+            RestResult res = restClient.execQuery("/");
+            if(res.getStatus() != 200)){
+                throw new DiagnosticException("Unrecoverable ES query error.");
```
pickypg:
```suggestion
                throw new DiagnosticException("Unable to fetch [GET /]");
```

Since this is in the `try` / `catch`, this info will go well with the error output.

src/main/java/com/elastic/support/rest/RestResult.java
```diff
@@ -85,4 +87,27 @@ public void toFile(String fileName){
             logger.log(SystemProperties.DIAG, "Error writing Response To OutputStream", e);
         }
     }
+
+    public static boolean isRetryable(int statusCode) {
+
+        if (statusCode == 400) {
+            logger.info("No data retrieved.");
```
pickypg:
```suggestion
            logger.info("[400] The request was rejected.");
```

src/main/java/com/elastic/support/rest/RestResult.java
```diff
@@ -85,4 +87,27 @@ public void toFile(String fileName){
...
+            return true;
+        } else if (statusCode == 401) {
+            logger.info("Authentication failure: invalid login credentials.");
```
pickypg:
```suggestion
            logger.info("[401] Authentication failure: invalid login credentials.");
```

src/main/java/com/elastic/support/rest/RestResult.java
```diff
@@ -85,4 +87,27 @@ public void toFile(String fileName){
...
+            return false;
+        } else if (statusCode == 403) {
+            logger.info("Authorization failure or invalid license.");
```
pickypg:
```suggestion
            logger.info("[403] Authorization failure or invalid license.");
```

src/main/java/com/elastic/support/rest/RestResult.java
```diff
@@ -85,4 +87,27 @@ public void toFile(String fileName){
...
+            return false;
+        } else if (statusCode == 404) {
+            logger.info("Endpoint does not exist.");
```
pickypg:
```suggestion
            logger.info("[404] Endpoint does not exist (is the expected plugin enabled?).");
```

src/main/java/com/elastic/support/rest/RestResult.java
```diff
@@ -85,4 +87,27 @@ public void toFile(String fileName){
...
+            return true;
+        } else if (statusCode >= 500 && statusCode < 600) {
+            logger.info("Undetermined server error.");
```
pickypg:
```suggestion
            logger.info("[" + statusCode + "] Undetermined server error.");
```

src/main/java/com/elastic/support/diagnostics/DiagnosticService.java
```diff
@@ -58,9 +57,13 @@ public void exec(DiagnosticInputs diagnosticInputs, DiagConfig diagConfig, Map<S
 
             dc.runDiagnostic(ctx, chain);
 
+            if (ctx.isDocker()){
+                logger.warn("Identified Docker installations - bypassing log collection and system calls.");
```
pickypg: This is really saying that log collection and system calls were _already_ bypassed, right? If so, perhaps we should make this more of a past tense comment or put it wherever that is detected?

nemonster: A few suggestions weren't necessarily warranted - for instance, the REST call is displayed immediately prior to it's submittal so putting it in the error message would be redundant. However I did go back to look this over and decided that I wasn't all that happy with it. So I reworked the error handling in general to separate out unanticipated exceptions from one I explicitly throw. And use the message in those DiagnosticExceptions as the logged message displayed. Also, keep in mind there are two types of logging here. The normal info, warn, error variety which the user sees and the custom level, which contains stack traces and is only written to the internal log contained in the diagnostic.

Issue #233, opened 2018-12-14
Invalid cookie header "Set-Cookie: AWSALB=xxxxxxxxxxxx; Expires=Fri, 21 Dec 2018 GMT; Path=/". Invalid 'expires' attribute: Fri, 21 Dec 2018 09:19:23 GMT


priyaprabhakar777:
Getting the error 
Invalid cookie header "Set-Cookie: AWSALB=xxxxxxxxxxxx; Expires=Fri, 21 Dec 2018 GMT; Path=/". Invalid 'expires' attribute: Fri, 21 Dec 2018 09:19:23 GMT
==========
We execute below

wget https://github.com/elastic/support-diagnostics/releases/download/6.4.4/support-diagnostics-6.4.4-dist.zip
cd support-diagnostics-6.4.4
./diagnostics.sh --host $ES_ENDPOINT -u $USERNAME -p --ssl

priyaprabhakar777:
More details on error
---------
17:19:24 WARN  Invalid cookie header: "Set-Cookie: AWSALB=xxxxxxxxxx; Expires=Fri, 21 Dec 2018 09:19:23 GMT; Path=/". Invalid 'expires' attribute: Fri, 21 Dec 2018 09:19:23 GMT
17:19:24 INFO  Checking for diagnostic version updates.
17:19:24 INFO  Currently running query: _alias?pretty
17:19:24 WARN  Invalid cookie header: "Set-Cookie: AWSALB=xxxxxxxxxx; Expires=Fri, 21 Dec 2018 09:19:24 GMT; Path=/". Invalid 'expires' attribute: Fri, 21 Dec 2018 09:19:24 GMT
17:19:24 INFO  Currently running query: _cat/allocation?v
17:19:25 WARN  Invalid cookie header: "Set-Cookie: AWSALB=xxxxxxxxxx; Expires=Fri, 21 Dec 2018 09:19:24 GMT; Path=/". Invalid 'expires' attribute: Fri, 21 Dec 2018 09:19:24 GMT
17:19:25 INFO  Currently running query: _cat/master?format=json
17:19:26 WARN  Invalid cookie header: "Set-Cookie: AWSALB=xxxxxxxxxx; Expires=Fri, 21 Dec 2018 09:19:25 GMT; Path=/". Invalid 'expires' attribute: Fri, 21 Dec 2018 09:19:25 GMT
[...]

PR #299, opened 2019-08-06
Update Cookie Spec

nemonster:
Updates cookie spec used to the more lenient 6265. Default will not change until next major version. Causes issues when run through some load balancers.
Closes: #233


nemonster:
Actually, I only updated this because now and then people disregard the instructions and run it through a load balancer. It's not a valid usage, but when it fails the person who's penalized the most is the support engineer because they don't get any kind of diag at all. It's now the recommended setting for HttpClient, and will be the default in v5. So I just decided to go with the path of least resistance.

Issue #302, opened 2019-08-12
Add option to scrub a single file instead of full diagnostic.


nemonster:
Sometimes customers only need to send one file, such as a log but are restricted to scrubbing the entire diagnostic, which can take some time. You can extract one file and create a new archive to feed it, but that option also extends the time it takes to get needed information.

PR #303, opened 2019-08-12
Fix 32 byte character limit on obfuscation.

### LINKED ISSUE #301: Scrubber throws `IndexOutOfBoundsException` for token matches greater than 32 characters
[1,3,0,1]
Using token-based scrubbing, if any token matches a string of greater than 32 characters, an exception is thrown and the target file gets omitted entirely from final scrubbed bundle. This seems to be down to the fact that replacement tokens are derived from a generated `UUID`, which is 32 bytes in length.

https://github.com/elastic/support-diagnostics/blob/master/src/main/java/com/elastic/support/scrub/ScrubProcessor.java#L162-L171

Simplest solution here would probably be just to pad the string with some arbitrary character, such as `=` as is convention for many other encoding schemes.

nemonster:
Fixed issue where Strings longer than 32 characters generated an error during token replacement.
Added option to obfuscate a single file rather than a full diagnostic archive.
Closes #301 , #302

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,14 +36,13 @@ As a first step the diagnostic will check the Github repo for the current releas
 * For air gapped environments this can be bypassed by adding `--bypassDiagVerify` to the command line.
 
 ## Usage - Simplest Case
-* Run the application via the diagnostics.sh or diagnostics.bat script. The host name or IP address used by the HTTP connector of the node is required.
+* Run the application via the diagnostics.sh or diagnostics.bat script. The host name or IP address used by the HTTP connector of the node is required. IP address will generally ensure the best result.
 * In order to assure that all artifacts are collected it is recommended that you run the tool with elevated privileges. This means sudo on Linux type platforms and via an Administor Prompt in Windows.
 * A hostname or IP address must now be specified via the `--host` parameter. This is due to the changes in default port binding what were introduced starting with version 2. You must supply this even if you are running as localhost.
 * The utility will use default listening port of `9200` if you do not specify one.
 * If the utility cannot find a running ES version for that host/port combination the utility will exit and you will need to run it again.
 * Input parameters may be specified in any order.
 * An archive with the format diagnostics-`<DateTimeStamp>`.tar.gz will be created in the utility directory. If you wish to specify a specific output folder you may do so by using the -o `<Full path to custom output folder>` option.
-
 
 #### Basic Usage Examples
     NOTE: Windows users use .\diagnostics.bat instead of ./diagnostics.sh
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -59,7 +58,7 @@ As a first step the diagnostic will check the Github repo for the current releas
 
 ## Using With Shield/Security
```
pickypg:
```suggestion
## Using With Security
```

README.md
```diff
@@ -59,7 +58,7 @@ As a first step the diagnostic will check the Github repo for the current releas
...
 * a truststore does not need to be specified - it's assumed you are running this against a node that you set up and if you didn't trust it you wouldn't be running this.
-* When using Shield authentication, do not specify a password and the `-p` option.  Using the `-p` option will bring up a prompt for you to type an obfuscated value that will not be displayed on the command history.
+* When using security authentication, do not specify a password and the `-p` option.  Using the `-p` option will bring up a prompt for you to type an obfuscated value that will not be displayed on the command history.
```
pickypg:
```suggestion
* When using authentication, do not specify both a password and the `-p` option.  Using the `-p` option will bring up a prompt for you to type an obfuscated value that will not be displayed on the command history.
```

src/main/java/com/elastic/support/scrub/ScrubProcessor.java
```diff
@@ -161,10 +158,21 @@ public String processIpv6Addresses(String input){
...
+      StringBuffer newToken = new StringBuffer();
+
+      int passes = 1;
+      if(len > 32){
+         passes = (len/32) + 1;
+      }
+
+      for(int i = 0; i < passes; i++){
+         newToken.append(
+                 ("" + (UUID.nameUUIDFromBytes(token.getBytes())).toString())
```
pickypg:
```suggestion
                 UUID.nameUUIDFromBytes(token.getBytes()).toString()
```

src/main/java/com/elastic/support/scrub/ScrubProcessor.java
```diff
@@ -161,10 +158,21 @@ public String processIpv6Addresses(String input){
 
    public String generateReplacementToken(String token, int len){
 
-      String replacement =
-         ("" + (UUID.nameUUIDFromBytes(token.getBytes())).toString())
-         .replaceAll("-", "")
-         .substring(0, len);
+      StringBuffer newToken = new StringBuffer();
```
pickypg:
```suggestion
      StringBuilder newToken = new StringBuilder();
```

src/main/java/com/elastic/support/util/ArchiveUtils.java
```diff
@@ -110,6 +109,13 @@ public void extractDiagnosticArchive(String sourceInput, String targetDir, boole
             String name = tae.getName();
             name = name.replace(baseArchivePath, "");
 
+            if(  name.contains("gc.log.") ||
+                    (name.contains(".log") && !name.equalsIgnoreCase("diagnostics.log") ) ||
```
pickypg: Will this still work on 7.x JSON-based logs?

nemonster: No, I haven't started on that yet. They won't be collected as of now. When I start on that I will need to address that here.

PR #304, opened 2019-08-19
Refactor to prepare for monitoring extraction

nemonster:
Split out configuration and input classes.
Split out service functionality.
Add formatter for JSON
Fix old SLF4J logging declaration
Clean up some unused imports and removed some unused variables.


src/main/java/com/elastic/support/ElasticClientService.java
```diff
@@ -0,0 +1,78 @@
...
+import com.elastic.support.util.SystemUtils;
+import org.apache.commons.lang3.StringUtils;
+
+public class ElasticClientService extends BaseService {
+
+
+    protected RestClient createGenericClient(BaseConfig config, ElasticClientInputs inputs) {
+        RestClientBuilder builder = new RestClientBuilder();
+
+        return builder
```
pickypg: If you provided a private version of this method that returned the `RestClientBuilder`, then both of them could share the "generic" builder and `createEsRestClient` could just add what's unique to it.

src/main/java/com/elastic/support/config/ElasticClientInputs.java
```diff
@@ -0,0 +1,145 @@
...
+            return false;
+        }
+
+        return true;
+
+    }
+
+    protected boolean isAuthValid(String user, String password) {
+
+        return (!StringUtils.isNotEmpty(user) || !StringUtils.isEmpty(password)) &&
```
pickypg: I had to re-read this a few times because of all of the `!` inversions happening (compounded by some of the methods themselves implicitly including `Not`). This may be a little easier to follow?

```java
boolean invalid =
  (StringUtils.isEmpty(user) && StringUtils.isNotEmpty(password)) ||
  (StringUtils.isNotEmpty(user) && StringUtils.isEmpty(password));

return !invalid;
```

PR #319, opened 2019-10-16
Fix for issue with --noVerify not working.

nemonster:
Needed to adjust the connection factory configuration - adding connection pooling broke bypassing verification in the last updates.

### DIFF
--- a/src/main/java/com/elastic/support/rest/RestClientBuilder.java
+++ b/src/main/java/com/elastic/support/rest/RestClientBuilder.java
@@ -29,7 +29,9 @@ import org.apache.http.ssl.SSLContextBuilder;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/elastic/support/rest/RestClientBuilder.java
```diff
@@ -259,5 +266,18 @@ public RestClient build() {
 
     }
 
+    /**
+     * This overrides any hostname mismatch in the certificate
+     */
+    private class BypassHostnameVerifier implements HostnameVerifier {
```
pickypg: Does this class actually get used?

nemonster: No, it was a test to see why the NoVerify was not working. It did actually used to be in there and was recently removed because Apache provided a built in class. Just left it in because I was in a hurry and didn't notice it until after I'd pushed it up.

Issue #277, opened 2019-05-10
[diag][tooling] Break out api's by sub-folder


aliciascott:
## Idea synopsis
Because there is so much good information in the diagnostic bundle now, would it be possible to list/group the api's by folder?

Example:
***ILM***
--ilm_explain.json
--ilm_policies.json
--ilm_status.json

***CCR***
--ccr_stats.json
--ccr_follower_info.json
--ccr_autofollow_patterns.json

***ML***
--ml_anomaly_detectors.json
--ml_datafeeds.json
--ml_stats.json

***Watcher***
--watcher*.json

***Rollup***
--rollup*.json

#### Use case
It would help to have the features not always used be listed in folders, but the general `cat_*` api's and `cluster` settings etc still in the base folder, for easier reading.

This is just an idea of course!

nemonster:
Added in 7.1.3.

PR #327, opened 2019-11-18
Configuration move to semver -centric version checks for queries.

### LINKED ISSUE #326: Enhance rest url maintenance
Move to semver library to allow embedding rule comparison in configuration.
Consolidate retry and file extension attributes into configuration.

nemonster:
Dependency and version updates.
Minor mods to logging config.
Separated out rest calls to elastic and logstash specific and changed versioning format from release based to call based.
Added test to validate format of rest configuration files at build time.
New classes for refactored rest entries.
Closes #326 , #277

### DIFF
--- a/src/main/java/com/elastic/support/rest/RestEntryBuilder.java
+++ b/src/main/java/com/elastic/support/rest/RestEntryBuilder.java
@@ -1,0 +1,62 @@
+package com.elastic.support.rest;
+
+import com.elastic.support.util.SystemProperties;
+import com.sun.org.apache.xpath.internal.operations.Bool;
+import com.vdurmont.semver4j.Semver;
+import jdk.internal.jshell.tool.resources.version;
+import org.apache.commons.lang3.ObjectUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import javax.swing.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class RestEntryBuilder {
+
+    private static final Logger logger = LogManager.getLogger(RestEntryBuilder.class);
+
+    Semver semver;
+    public RestEntryBuilder(String version){
+        semver = new Semver(version);
+    }
+
+    public RestEntry build(Map.Entry<String, Object> entry){
+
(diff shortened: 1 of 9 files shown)


src/main/java/com/elastic/support/rest/RestEntryBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.commons.lang3.ObjectUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import javax.swing.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class RestEntryBuilder {
```
pickypg: Nitpick: This is more of a `Factory` than a `Builder`. It also doesn't seem to be used, but I'm guessing that's for the next PR?

src/main/resources/logstash-rest.yml
```diff
@@ -0,0 +1,38 @@
...
+
+logstash_version:
+  extension: ".json"
+  versions:
+    "> 0.0.0": "/?pretty"
+
+logstash_node:
+  extension: ".json"
+  versions:
+    "> 0.0.0": "/_node?pretty"
```
pickypg: These versions aren't true, right? I forget what version we actually added APIs to Logstash, but I didn't think it was until at least 2.x? Obviously with the `version` check API we can't know until we try.

src/main/java/com/elastic/support/rest/RestEntryBuilder.java
nemonster: True, it changed substantially at one point and I never renamed it. Next PR will have the implementation. At some point I'll probably rename it but I had other higher priorities.

src/main/resources/logstash-rest.yml
nemonster: It was 5.x, but if you look, the current version has no checks in code, just instructions not to use is before v.5. So I kept it consistent with what was there for now. The fact that nobody has ever raised an issue on that probably means it never got tried on an earlier version, At some point I'll make another pass through to see if there's any differentiation that requires a version check?

PR #330, opened 2019-11-21
Refactored command chain

nemonster:
Moved pathways out of configuration into code


src/main/java/com/elastic/support/diagnostics/chain/DiagnosticChainExec.java
```diff
@@ -2,22 +2,42 @@
...
 
     private static Logger logger = LogManager.getLogger(DiagnosticChainExec.class);
 
-    public void runDiagnostic(DiagnosticContext context, List<String> chain) {
+    public static void runDiagnostic(DiagnosticContext context, String type) {
 
         try {
-            Chain diagnostic = new Chain(chain);
-            diagnostic.execute(context);
+            if("standard".equals(type) || "remote".equals(type)) {
```
pickypg: Seems like this could be a `switch` statement with `"standard"` falling through to `"remote"`? Unless order of the commands is important.

Also, it seems like this is missing an `else` caveat to catch an invalid type making it through.

src/main/java/com/elastic/support/diagnostics/chain/DiagnosticChainExec.java
```diff
@@ -2,22 +2,42 @@
...
+                new GenerateManifestCmd().execute(context);
+                new DiagVersionCheckCmd().execute(context);
+                new VersionCheckCmd().execute(context);
+                new UserRoleCheckCmd().execute(context);
+                new RunClusterQueriesCmd().execute(context);
+            }
+
+            if("standard".equals(type)){
```
pickypg: If it's not worth a `switch` statement, then this should probably be nested under the above `if`, since it's implicitly `false` if that's `false` and 50/50 `true` if it's `true`.

Then, you can make the `"logstash"` check an `else if` and add an `else`.

src/main/java/com/elastic/support/diagnostics/chain/DiagnosticChainExec.java
nemonster: This is just an interim step to change it over from the configuration file. Type options are going to be changing anyway before the release.

PR #338, opened 2019-11-26
Fix Clustername Modification

nemonster:
* Fixed issue with cluster name not being updated in imported indices
* Fixed validation issue with input clustername
* Modified Service Objects in export/import to be threadsafe.

### DIFF
--- a/src/main/java/com/elastic/support/monitoring/MonitoringExportApp.java
+++ b/src/main/java/com/elastic/support/monitoring/MonitoringExportApp.java
@@ -17,7 +17,6 @@ public class MonitoringExportApp {
             System.exit(0);
         }
 
-        MonitoringExportService service = new MonitoringExportService(monitoringExportInputs);
-        service.execExtract();
+        new MonitoringExportService().execExtract(monitoringExportInputs);
     }
 }
--- a/src/main/java/com/elastic/support/monitoring/MonitoringExportService.java
+++ b/src/main/java/com/elastic/support/monitoring/MonitoringExportService.java
@@ -28,24 +28,24 @@ public class MonitoringExportService extends ElasticRestClientService {
 
     private Logger logger = LogManager.getLogger(MonitoringExportService.class);
     private static final String SCROLL_ID = "{ \"scroll_id\" : \"{{scrollId}}\" }";
-    private RestClient client;
-    private MonitoringExportConfig config;
-    private MonitoringExportInputs inputs;
-    private String tempDir = SystemProperties.userDir + SystemProperties.fileSeparator + Constants.MONITORING_DIR;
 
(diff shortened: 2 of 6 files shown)


src/main/java/com/elastic/support/monitoring/MonitoringExportApp.java
```diff
@@ -17,7 +17,6 @@ public static void main(String[] args) {
             System.exit(0);
         }
 
-        MonitoringExportService service = new MonitoringExportService(monitoringExportInputs);
-        service.execExtract();
+        new MonitoringExportService().execExtract(monitoringExportInputs);
```
pickypg: Seems like this method makes more sense as a `static` one now that its only attribute is the logger?

src/main/java/com/elastic/support/monitoring/MonitoringExportService.java
```diff
@@ -28,24 +28,24 @@
...
 
-    public void execExtract() {
+        RestClient client = null;
+        Map configMap = JsonYamlUtils.readYamlFromClasspath(Constants.DIAG_CONFIG, true);
+        MonitoringExportConfig config = new MonitoringExportConfig(configMap);
 
         try {
+            client = createEsRestClient(config, inputs);
+
+            String tempDir = "";
```
pickypg:
```suggestion
            String tempDir = SystemProperties.fileSeparator + Constants.MONITORING_DIR;
```

src/main/java/com/elastic/support/monitoring/MonitoringExportService.java
```diff
@@ -28,24 +28,24 @@
...
+            if(StringUtils.isNotEmpty(inputs.getOutputDir())){
+                tempDir = inputs.getOutputDir() + SystemProperties.fileSeparator + Constants.MONITORING_DIR;;
```
pickypg:
```suggestion
                tempDir = inputs.getOutputDir() + tempDir;
```

src/main/java/com/elastic/support/monitoring/MonitoringExportService.java
```diff
@@ -28,24 +28,24 @@
...
+            }
+            else{
+                tempDir = SystemProperties.userDir + SystemProperties.fileSeparator + Constants.MONITORING_DIR;
```
pickypg:
```suggestion
                tempDir = SystemProperties.userDir + tempDir;
```

src/main/java/com/elastic/support/monitoring/MonitoringExportApp.java
nemonster: You could actually say that about a number of the current classes. But I don't want to do that yet - I still may change my mind.

PR #351, opened 2019-12-16
Added one-liner to cd to the script directory

robin13:


nemonster:
@robin13 
Before I review the actual change:
* Exactly why do you want to do this? Exactly what problem is it solving? Please be as detailed as possible.
* I don' see a corresponding update to the readme - if you change how something works it needs to be documented.  
* There are two execution scripts, one for 'nix and one for Windows. For something at the level of the basic script run, I'm not that fond of having two different platform specific actions. This also relates to the the documentation item above.

robin13:
> Exactly why do you want to do this? Exactly what problem is it solving? Please be as detailed as possible.

The problem was a user who ran the script (in a docker container) without changing to the script directory first.  It failed with a cryptic (for non java developers) error indicating that it could not find the necessary libraries.  By always changing to the script directory first (regardless of if you are already in it or not) you avoid that failure profile.

> I don' see a corresponding update to the readme - if you change how something works it needs to be documented.

For the user using it as they had before (running the script from the directory) nothing has changed on how it works.  I don't see that a bug-fix needs an entry in the README?

I guess it could be argued that since the readme contains the line:

> * Switch to the diagnostics distribution directory.

This fix is not necessary as running it from elsewhere is not running it as documented... but as some people have, and some people have had a problem as a result ("If a new user has a bad time, it's a bug") I would think making usage easier would always be good.

nemonster:
@robin13 I'm adding extended functionality to allow running outside the install directory but I'm doing it differently and for both 'nix and Windows scripts so closing this one.

PR #359, opened 2020-02-03
Diagnostic Refactor

nemonster:
- Adds capability to get system logs and calls from a remote node.
- New set of type designations where elasticsearch and logstash share a set of conventions
    local - run against process on that host
    remote - run against process on a different host
    api - no system calls or logs, only REST calls 
- Interactive mode to step user through options
- host no longer required parameter. Defaults to localhost.
- All batch password prompts modified to be collected from the same methods as the interactive so each needs to be done via a specific option as in the Elastic password . This can be overridden to send them in via the command line but this information is not displayed in the help.
- For REST calls, diag will try to determine the current master and whether that master is configured for HTTP access. If so, the master will be queried rather than the target node. Target node will have system calls and logs collected.
- new version of license call compatible with Enterprise license added.
- Removed execution sequences from configuration to allow conditional processing.
- Removed option to specify a specific docker id. If a docker image is encountered, all running instances will be collected.
- Fixed issue with Logstash execution against remote process or Docker image failed.
- Fixed issue with running export and import with defaults rather than specifying all options explicitly.
- Modified check diagnostic to allow easier running under debugger.
- Added generated exception messages from REST call to displayed error line in addition to internally logging the exception.


src/main/java/com/elastic/support/BaseInputs.java
```diff
@@ -2,57 +2,147 @@
...
+
+    public List<String> validateFile(String val) {
+        if (StringUtils.isEmpty(val.trim())) {
+            return null;
+        }
+
+        File file = new File(val);
+
+        if (!file.exists()) {
+            return Collections.singletonList("Specified file could not be located.");
```
pickypg: Probably worth putting the input value with `""` or `[]` wrapping it so that it's easier to debug (e.g., on Windows where paths sometimes have spaces, so if they forgot to use quotes it would be a little more apparent here).

Same would probably help for all of the input errors, but especially for the file ones.

src/main/java/com/elastic/support/BaseInputs.java
```diff
@@ -2,57 +2,147 @@
...
+            return Collections.singletonList("Specified required file location could not be located.");
+        }
+
+        return null;
+
     }
 
     public String toString(){
-        return "Output Directory: " + outputDir + SystemProperties.lineSeparator;
+       return "outputDir: " + outputDir + ",";
```
pickypg: Seems like it should be the job of the caller to add the `","` if it wants it.

src/main/java/com/elastic/support/BaseService.java
```diff
@@ -77,9 +76,8 @@ public void createArchive(String tempDir) {
             archiveUtils.createArchive(tempDir, archiveFilename);
 
         } catch (Exception ioe) {
-            logger.error("Couldn't create archive. {}", ioe);
+            logger.info("Couldn't create archive. {}", ioe);
```
pickypg: Not that I _really_ care, but this seems like a solid reason for an error versus info?

src/main/java/com/elastic/support/diagnostics/DiagnosticInputs.java
```diff
@@ -3,92 +3,265 @@
...
         }
+        return null;
+    }
 
-        return true;
-
+    public List<String> validateDiagType(String val) {
+        List<String> types = Arrays.asList(diagnosticTypeValues);
+        if (!types.contains(val)) {
+            return Collections.singletonList(val + " was not a valid diagnostype type. Enter --help to see valid choices");
```
pickypg:
```suggestion
            return Collections.singletonList(val + " was not a valid diagnostic type. Enter --help to see valid choices");
```

remote-diagnostics/cat/cat_master.txt
```diff
@@ -0,0 +1 @@
+[{"id":"4Kmf8EYORYas5rHMd7fTuw","host":"172.18.0.3","ip":"172.18.0.3","node":"4Kmf8EY"}]
```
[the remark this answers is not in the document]
nemonster: Yes. At least when you run GET /_cat/master?format=json.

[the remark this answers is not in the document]
pickypg: That makes sense, but shouldn't we call the file `cat_master.json` then?

src/main/java/com/elastic/support/BaseConfig.java
```diff
@@ -1,35 +1,57 @@
...
+    public String diagnosticVersion;
+    public int connectionTimeout;
+    public int connectionRequestTimeout;
+    public int socketTimeout;
+    public int maxTotalConn;
+    public int maxConnPerRoute;
 
-    public BaseConfig() {
-    }
+    public String diagReleaseHost = "api.github.com";
```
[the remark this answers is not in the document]
nemonster: I actually plan to either remove these from the configuration or comment them. The only time they'd need to be in the config is if something changed prior to a release and you needed a way to modify the config as a workaround. But you do then get that "get out of jail free card". Had intended to make a final pass through and possibly make some additional changes to the config in a future change.

And on that note, if you make it final it seems that stops you from setting it via config. And in this particular case, it doesn't hurt I guess, but what's going to change it?  What set of circumstances would cause somebody to do that? Usually if I can't easily think of one in 15 minutes or so I don't worry about it.

src/main/java/com/elastic/support/BaseInputs.java
```diff
@@ -2,57 +2,147 @@
...
+            .withInputTrimming(true);
+    protected BooleanInputReader standardBooleanReader = ResourceCache.textIO.newBooleanInputReader();
+    protected StringInputReader  standardPasswordReader = ResourceCache.textIO.newStringInputReader()
+            .withInputMasking(true)
+            .withInputTrimming(true)
+            .withMinLength(0);
+    protected StringInputReader standardFileReader = ResourceCache.textIO.newStringInputReader()
+            .withInputTrimming(true)
+            .withValueChecker((String val, String propname) -> validateFile(val));
+    // Input Readers
```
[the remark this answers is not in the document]
nemonster: I guess, it's just a shortcut marker for going fwd/rev. Had considered taking them out as well.

src/main/java/com/elastic/support/BaseService.java
nemonster: If you go back and check the convention, all of the log statements that aren't logger.log are info. That's because I use a custom appender level that writes everything to to the bundled log file and only some things to the console. None of which even display the level now. So what level you log it at is lost on whoever's looking at the console. Until the appender gets configured though, i still need some default configured behavior to output messages. It's easy to get one level, a lot more work to get more than that. With no visible payoff.

src/main/java/com/elastic/support/BaseInputs.java
nemonster: That one is kind of trivial, but this is just the base class. It's never used by itself. I did go back and check the diagnostic inputs and when I regenerated the toString I lost the call the the superclass. So I'm fixing the DiagnosticInputs as well.

src/main/java/com/elastic/support/diagnostics/DiagnosticInputs.java
nemonster: Done.

src/main/java/com/elastic/support/BaseInputs.java
nemonster: If they forget to use quotes the command line parsing utility I use will error out because it will look like a dangling argument that wasn't configured. And if they do put the quotes in there the input library will preserve them. IOW if you put in C:\Users\Greg Nieman\output it won't work. " C:\Users\Greg Nieman\output" _will_ work, and the quotes will carry over into the variable.

But given how the error will occur, it probably warrants emphasizing this in the doc changes I'm working on, even if I am skeptical anyone will pay any attention.

remote-diagnostics/cat/cat_master.txt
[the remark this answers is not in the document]
nemonster: OK, I see what you're getting at. I just didn't connect with the extension. Sure, I had always just defaulted all the cat calls to text. Will probably change the name of the call and the file though.

PR #368, opened 2020-02-11
Allow run from outside install directory

nemonster:
Allow run of utility from directory outside the unzipped diagnostic directory.


nemonster:
@pickypg - kind of waiting on this.

PR #369, opened 2020-02-13
Docker Container Awareness

nemonster:
Check for docker artifacts in base input class.
Resets output dir to default volume
Removes local run options.
Docker build script
Dockerfile
Sample docker container run
Fix for GH option check.

### DIFF
--- a/src/main/java/com/elastic/support/diagnostics/DiagnosticInputs.java
+++ b/src/main/java/com/elastic/support/diagnostics/DiagnosticInputs.java
@@ -5,6 +5,7 @@ import com.elastic.support.Constants;
 import com.elastic.support.rest.ElasticRestClientInputs;
 import com.elastic.support.util.ResourceCache;
 import com.elastic.support.util.SystemProperties;
+import com.elastic.support.util.SystemUtils;
 import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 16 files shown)


src/main/java/com/elastic/support/monitoring/MonitoringExportInputs.java
```diff
@@ -46,6 +46,10 @@
     public String queryStartDate;
     public String queryEndDate;
 
+    public MonitoringExportInputs(){
```
pickypg: Default constructor with `super();` is even more implied.

src/main/java/com/elastic/support/monitoring/MonitoringImportInputs.java
```diff
@@ -37,6 +37,10 @@
 
     // End Input Readers
 
+    public MonitoringImportInputs(){
```
pickypg: Ditto.

src/main/java/com/elastic/support/rest/ElasticRestClientInputs.java
```diff
@@ -86,6 +86,13 @@
 
     Logger logger = LogManager.getLogger(ElasticRestClientInputs.class);
 
+    public ElasticRestClientInputs(){
+        super();
```
pickypg: Ditto.

src/main/java/com/elastic/support/scrub/ScrubInputs.java
```diff
@@ -45,6 +45,9 @@ public void setConfigFile(String configFile) {
 
     // End Input Fields
 
+    public ScrubInputs(){
```
pickypg: Ditto.

src/main/java/com/elastic/support/diagnostics/DiagnosticInputs.java
```diff
@@ -106,28 +122,35 @@
     public int remotePort = 22;
     // End Input Fields
 
+    String[] typeEntries;
+
+    public DiagnosticInputs(){
+        super();
```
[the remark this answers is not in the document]
nemonster: Here's the thing though. I added these after when testing my Docker container tests were failing. Because of the output directory not being correctly reset in the BaseInputs constructor. For some reason it wasn't firing. I came back to it the next day and added the super in the DiagnosticInputs and things seemed to be working. So that was a pre-emptive insurance policy to get it running. After revisiting it, I removed the super and it's firing off for some reason. That part of the code doesn't change much, so I'm at kind of a loss on this one.

PR #373, opened 2020-02-18
Scrub Interactive

nemonster:
Added interactive mode to scrub
Cleaned up extraneous debug code.

### DIFF
--- a/src/main/java/com/elastic/support/scrub/ScrubApp.java
+++ b/src/main/java/com/elastic/support/scrub/ScrubApp.java
@@ -21,10 +21,7 @@ public class ScrubApp {
         try {
             ScrubInputs scrubInputs = new ScrubInputs();
             if(args.length == 0){
-                logger.info("{}{}{}",
-                        "Command line options can be displayed with the --help arguemnt.",
-                        SystemProperties.lineSeparator,
-                        "Entering interactive mode.");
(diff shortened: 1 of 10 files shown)


src/main/java/com/elastic/support/scrub/ScrubApp.java
```diff
@@ -17,19 +20,16 @@ public static void main(String[] args) {
...
             if(args.length == 0){
-                scrubInputs.interactive = true;
-            }
-            List<String> errors = scrubInputs.parseInputs(args);
-
-            if( args.length == 0 || scrubInputs.interactive){
-                // Create a new input object so we out clean
-                scrubInputs = new ScrubInputs();
+                logger.info("{}{}{}",
+                        "Command line options can be displayed with the --help arguemnt.",
```
pickypg: I wonder if we it might be easier to read this if it flips the order of the text:

```
Entering interactive mode. Command line options can be displayed with the --help arguemnt.
```

That way people see that it's entering interactive mode and can stop reading (rather than see the first line, then stop reading before the second). And you can do away with the line separator from appearing in the logs.

pickypg: I just noticed there's a typo: `arguemnt` -> `argument`

PR #374, opened 2020-02-18
Check for source distribution

### LINKED ISSUE #203: Error: Could not find or load main class com.elastic.support.diagnostics.DiagnosticApp
Im having this issue :

coltapps@uklvadapp092[DEV][scripts] $ sudo find / -name com.elastic.support.diagnostics.DiagnosticApp
coltapps@uklvadapp092[DEV][scripts] $ ./diagnostics.sh --host 10.198.39.5
Using /usr/bin/java as Java Runtime
Error: Could not find or load main class com.elastic.support.diagnostics.DiagnosticApp
coltapps@uklvadapp092[DEV][scripts] $ **Error: Could not find or load main class com.elastic.support.diagnostics.DiagnosticApp**
-bash: Error:: command not found
coltapps@uklvadapp092[DEV][scripts] $ pwd
/home/coltapps/support-diagnostics-6.3/scripts
coltapps@uklvadapp092[DEV][scripts] $

nemonster:
Displays error message and exits if it cannot find the lib directory created when building the runtime distribution.
closes #203

### DIFF
--- a/scripts/export-monitoring.bat
+++ b/scripts/export-monitoring.bat
@@ -3,6 +3,13 @@ setlocal enabledelayedexpansion
 
 SET scriptpath=%~dp0
 SET diagpath=%scriptpath:~0,-1%
+SET libpath=%diagpath%\lib\NUL
+
+IF NOT EXIST %libpath% (
+    ECHO "Runtimes library does not exist - make sure you are running the "
+    ECHO "archive with 'dist' in the name, not the one labeled: 'source'."
+    EXIT
+)
 
 set JAVA_EXEC=java
 if not defined JAVA_HOME (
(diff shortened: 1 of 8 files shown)


scripts/export-monitoring.bat
```diff
@@ -3,6 +3,13 @@ setlocal enabledelayedexpansion
 
 SET scriptpath=%~dp0
 SET diagpath=%scriptpath:~0,-1%
+SET libpath=%diagpath%\lib\NUL
```
pickypg: Do these all need the `\NUL` terminator? You're consistently only doing it in the `.bat` files, but I'm just not used to seeing it and wanted to call out in case it was leftover from something.

nemonster: It's a Windows thing. When checking on stackoverflow there appeared to be some controversy over whether the NUL was needed. Evidently in some cases Windows won't check for a folder, only a file and this forces the directory to a file so it always works. It does work without it in my testing but I left it in there in case it's tied to some Windows version or update.

PR #398, opened 2020-04-08
fix documentation about parameter to ask for keystore password

sehdorn:
Hi guys,

the documentation of how to provide a password for a keystore provided, in case of using pki-authentication, points to a parameter called:
pkiKeyPass
which is not existing in:
https://github.com/elastic/support-diagnostics/blob/master/src/main/java/com/elastic/support/rest/ElasticRestClientInputs.java#L62
So i guess, this fixes the documentation 
Worked for me, at least

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@
  </tr>
  
  <tr>
-   <td width="20%" align="left" valign="top">--pkiKeyPass</td>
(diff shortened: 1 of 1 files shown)


nemonster:
Restructured the interactive inputs via a different PR. Fixed the readme in that so this is no longer needed. Closing.

PR #411, opened 2020-05-14
Allow for no scrub config file

AndyHunt66:
If you want to use the default scrub configuration, the interactive text tells you you can just hit enter to use the default.

However, if you do just hit enter, it complains that you have to provide a file:
```
Invalid value.
Input file is required.
```

This change uses the optional file validation instead of the required file validation


nemonster:
I'm actually going to change the structure of the configuration files so you won't input this. It will just be there in a _config_ directory to edit. So given that, I'm going to just close this one in favor of that solution.

Issue #416, opened 2020-05-15
Collect hidden indices stats and names


lucabelluccini:
Following the introduction of [hidden indices](https://github.com/elastic/elasticsearch/pull/50452), it is necessary to grab all the indices (including the hidden ones).

It will require the parameter `expand_wildcards=all` in most of the APIs which target `_all`/`*` indices.

PR #417, opened 2020-05-18
Support hidden indices

lucabelluccini:
This PR should cover most of the APIs returning indices to include hidden indices.
Solves https://github.com/elastic/support-diagnostics/issues/416

### DIFF
--- a/src/main/resources/elastic-rest.yml
+++ b/src/main/resources/elastic-rest.yml
@@ -51,7 +51,7 @@ cat_indices:
   subdir: "cat"
   versions:
     ">= 0.9.0 < 5.1.1": "/_cat/indices?v"
-    ">= 5.1.1": "/_cat/indices?v&s=index"
+    ">= 5.1.1": "/_cat/indices?v&s=index&expand_wildcards=all"
 
 cat_master:
   extension: ".txt"
@@ -176,7 +176,7 @@ fielddata_stats:
 indices_stats:
   retry: true
   versions:
-    ">= 0.9.0": "/_stats?level=shards&pretty&human"
+    ">= 0.9.0": "/_stats?level=shards&pretty&human&expand_wildcards=all"
 
 licenses:
   versions:
@@ -187,7 +187,7 @@ licenses:
 
 mapping:
   versions:
-    ">= 0.9.0": "/_mapping?pretty"
+    ">= 0.9.0": "/_mapping?pretty&expand_wildcards=all"
 
 master:
   versions:
@@ -231,7 +231,7 @@ segments:
 settings:
   retry: true
   versions:
-    ">= 0.9.0": "/_settings?pretty&human"
+    ">= 0.9.0": "/_settings?pretty&human&expand_wildcards=all"
 
 shard_stores:
   versions:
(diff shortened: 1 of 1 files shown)


src/main/resources/elastic-rest.yml
```diff
@@ -51,7 +51,7 @@ cat_indices:
   subdir: "cat"
   versions:
     ">= 0.9.0 < 5.1.1": "/_cat/indices?v"
-    ">= 5.1.1": "/_cat/indices?v&s=index"
+    ">= 5.1.1": "/_cat/indices?v&s=index&expand_wildcards=all"
```
lucabelluccini, line 54:
```suggestion
    ">= 5.1.1 < 7.7.0": "/_cat/indices?v&s=index"
    ">= 7.7.0": "/_cat/indices?v&s=index&expand_wildcards=all"
```

src/main/resources/elastic-rest.yml
```diff
@@ -176,7 +176,7 @@ fielddata_stats:
 indices_stats:
   retry: true
   versions:
-    ">= 0.9.0": "/_stats?level=shards&pretty&human"
+    ">= 0.9.0": "/_stats?level=shards&pretty&human&expand_wildcards=all"
```
lucabelluccini, line 179:
```suggestion
    ">= 0.9.0 < 7.7.0": "/_stats?level=shards&pretty&human"
    ">= 7.7.0": "/_stats?level=shards&pretty&human&expand_wildcards=all"
```

src/main/resources/elastic-rest.yml
```diff
@@ -187,7 +189,7 @@ licenses:
 
 mapping:
   versions:
-    ">= 0.9.0": "/_mapping?pretty"
+    ">= 0.9.0": "/_mapping?pretty&expand_wildcards=all"
```
lucabelluccini, line 192:
```suggestion
    ">= 0.9.0 < 7.7.0": "/_mapping?pretty"
    ">= 7.7.0": "/_mapping?pretty&expand_wildcards=all"
```

src/main/resources/elastic-rest.yml
```diff
@@ -231,7 +234,7 @@ segments:
 settings:
   retry: true
   versions:
-    ">= 0.9.0": "/_settings?pretty&human"
+    ">= 0.9.0": "/_settings?pretty&human&expand_wildcards=all"
```
lucabelluccini, line 237:
```suggestion
    ">= 0.9.0 < 7.7.0": "/_settings?pretty&human"
    ">= 7.7.0": "/_settings?pretty&human&expand_wildcards=all"
```

lucabelluccini:
Hello @nemonster 

I've updated the requests to respect the minimum versions/ranges.
In theory, the `expand_wildcards` setting was present since several versions, but it is better to stay safe.

> IIRC, any call that will take index as a parameter needs this. There are more of those calls than are represented here, _cat segments for instance. I think it warrants another pass through to ensure that everything which meets that criterion (at least if you wish to see the content of the hidden indices) is in there.

I've tested those and those requests are not affected by the the hidden indices (at least this is what I've witnessed on my 7.7.0 cluster).
- The `cat` shard API returns all the shards (Even those from hidden indices)
- The `cat` segment API returns only the shards of the non-hidden indices and passing the parameter. The `expand_wildcard` is not applicable and the only workaround I've found is: `GET _cat/segments/*,.*`

PR #423, opened 2020-06-03
-i on monitoring import causes issue

predogma:
Examples should be updated to use --input and --clusterName
Currently the use -i, -clusterName

### DIFF
--- a/README.md
+++ b/README.md
@@ -682,19 +682,19 @@
  Uses the default cluster_id, index_name:  
  
  ```$xslt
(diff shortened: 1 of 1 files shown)


nemonster:
I've been working on this functionality and have now done a couple of things.
* Since there's an -o and --output that's been there a while I decided I should keep this consistent so I'm changing it back to -i and --input. That will be reflected in the changed documentation.
* I decided to make that parameter clustername all in lower case just to make it easier to type. Again, will make sure that's reflected in the examples.
Thanks for those catches though.

PR #445, opened 2020-08-31
Update README.md

calebchaney:
close out open paren

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@
  
  # Support Diagnostics Utility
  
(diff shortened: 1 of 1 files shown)


calebchaney:
Since it's not obvious in the diff,

>It is compatible with all versions of Elasticsearch (including alpha, beta and release candidates, and for Logstash versions greater than v5.

to

>It is compatible with all versions of Elasticsearch (including alpha, beta and release candidates), and for Logstash versions greater than v5.

closing parenthesis after "candidates."

PR #446, opened 2020-09-23
Fix downloading-and-installing readme section

jguay:
I no longer see on top a section called "<some number> Releases" the readme refers to, simplifying by just specifying the link to the latest release


jpcarey (approved):
LGTM

PR #447, opened 2020-10-02
Update README.md

jguay:
--password no longer allows to pass plain text but readme still mentioned this was possible : --passwordText now allows this feature which is already documented

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@
  
  <tr>
    <td width="20%" align="left" valign="top">-p<br/>--password</td>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -201,7 +201,7 @@
  
  <tr>
    <td width="20%" align="left" valign="top">-p<br/>--password</td>
-   <td width="50%" align="left" valign="top">Generates obfuscated prompt for the elasticsearch password. Passing of a plain text password for automated processes is possible but not encouraged given it cannot be concealed from the history. See documentation for details. All other password prompts function in a similar fashion. </td>
+   <td width="50%" align="left" valign="top">Generates obfuscated prompt for the elasticsearch password. All other password prompts function in a similar fashion. </td>
```
pickypg, line 204: I wonder if we should mention `--passwordText` in this for those looking for it.

PR #470, opened 2021-02-16
new Kibana type

CamiloSierraH:
Hi everyone, here there are a few changes that will add a new type to the diagnostic.
You can see more details about this new Kibana type [here](https://github.com/elastic/support-ops/issues/2872).

Before you start your review, please allow me give you some context. In the Diagnostic current version 8.1.3:
- There are 0 test executed (those who failed are not executed),
- The code has not a mocking Framework,
- There is no doc in class or functions.
I give you some context because this is a minor PR, with this PR we will not move from 0% coverage to 100%, neither I have changed the structure of the code, I used the tools defined for Elasticsearch or Logstash, so we can create the new Kibana type.
I fixed the test that failed, I created new tests, and added docs for the function I created or used.

Why we don't have a framework for mocking class? I don't know, I was thinking to add Mockito but I think this is a decision that need to be made by Plumbers, as this can have an impact on the future of the project.

How to test the edge cases:
- The API can be tested easily with Elastic Kibana cloud,
- The remote I have a Kibana installed in a raspberry pi, 
- For Windows I created and GCP server, if you want to use it please let me know (the image has everything you need already installed).

On top of this, I have found an issue on Kibana, the actions API in Kibana is not working if you have installed Kibana using the zip/tar version, I have discussed with the Kibana team I need to open the issue.
And the /api/detection_engine/privileges API sent a 500 error if you have a kibana with no detections.
Both errors are logged but both don't have any impact on how the diagnostic works, all the other APIs and system calls, or collection of logs are executed.

### DIFF
--- a/src/test/java/com/elastic/support/diagnostics/commands/TestCheckKibanaVersion.java
+++ b/src/test/java/com/elastic/support/diagnostics/commands/TestCheckKibanaVersion.java
@@ -1,0 +1,151 @@
+package com.elastic.support.diagnostics;
+
+import com.elastic.support.Constants;
+import com.elastic.support.util.*;
+import org.junit.jupiter.api.Test;
+import com.elastic.support.diagnostics.commands.CheckKibanaVersion;
+import com.elastic.support.diagnostics.chain.DiagnosticContext;
+import org.mockserver.integration.ClientAndServer;
+import org.mockserver.matchers.Times;
+import org.mockserver.model.Header;
+import org.mockserver.model.Parameter;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.junit.jupiter.api.*;
+import org.apache.http.client.methods.*;
+import org.apache.http.impl.client.*;
+import com.elastic.support.rest.RestClient;
+import com.elastic.support.rest.RestEntry;
+import com.vdurmont.semver4j.Semver;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockserver.integration.ClientAndServer.startClientAndServer;
+import static org.mockserver.model.HttpRequest.request;
+import static org.mockserver.model.HttpResponse.response;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class TestCheckKibanaVersion {
+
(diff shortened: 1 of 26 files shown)


src/test/java/com/elastic/support/diagnostics/commands/TestCheckKibanaVersion.java
```diff
@@ -0,0 +1,151 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockserver.integration.ClientAndServer.startClientAndServer;
+import static org.mockserver.model.HttpRequest.request;
+import static org.mockserver.model.HttpResponse.response;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class TestCheckKibanaVersion {
+
+	private static final Logger logger = LogManager.getLogger(RestClient.class);
```
insukcho, line 29: Can we reformat this class for removing broken indention? Only this file has several places with different indention.

insukcho:
About automate testing, I think we can follow Kibana current functional testing framework from here:
https://www.elastic.co/guide/en/kibana/master/development-tests.html#_writing_functional_tests

We are using [WebDriver](https://www.w3.org/TR/webdriver1/) to test APIs without any Kibana installation and using [Mocha](https://mochajs.org/) as a test framework we need to write a script for the test.

pickypg:
> Why we don't have a framework for mocking class?

Some engineers are not fans of mocking and others are. I'm a big fan of it. The downside to mocking is that you can make "fake" tests that pass by simply returning junk that the mocked service never would, so then things look good even though it's really broken.

> About automate testing, I think we can follow Kibana current functional testing framework from here:

We cannot use WebDriver here because that is a test driver that pushes changes via Javascript (not Java) to be run in the browser, which the diagnostic does not use.

We should _at least_ be using JUnit for unit tests and, relative to mocking, something like Mockito (I would actually choose Mockito).

pickypg:
This PR seems to have commits that are not from this PR. Are you able to remove those?

CamiloSierraH:
I will close this PR, There are some commits that are not suppose to be here, I guess there was an issue when updating the fork. I created a new fork and cherry-pick my commits.

PR #471, opened 2021-02-18
new Kibana type

CamiloSierraH:
Hi everyone, here there are a few changes that will add a new type to the diagnostic.
You can see more details about this new Kibana type here.

Before you start your review, please allow me give you some context. In the Diagnostic current version 8.1.3:

-  There are 0 test executed (those who failed are not executed),
-   The code has not a mocking Framework,
-    There is no doc in class or functions.
-    I give you some context because this is a minor PR, with this PR we will not move from 0% coverage to 100%, neither I have changed the structure of the code, I used the tools defined for Elasticsearch or Logstash, so we can create the new Kibana type.
-    I fixed the test that failed, I created new tests, and added docs for the function I created or used.

Why we don't have a framework for mocking class? I don't know, I was thinking to add Mockito but I think this is a decision that need to be made by Plumbers, as this can have an impact on the future of the project.

How to test the edge cases:

-   The API can be tested easily with Elastic Kibana cloud,
-   The remote I have a Kibana installed in a raspberry pi,
-    For Windows I created and GCP server, if you want to use it please let me know (the image has everything you need already installed).

On top of this, I have found an issue on Kibana, the actions API in Kibana is not working if you have installed Kibana using the zip/tar version, I have discussed with the Kibana team I need to open the issue.
And the /api/detection_engine/privileges API sent a 500 error if you have a kibana instance with no detections defined.
Both errors are logged but both don't have any impact on how the diagnostic works, all the other APIs and system calls, or collection of logs are executed.


kostasb:
@elastic/support-ops Can you please prioritize to review this PR? We need to get Kibana API diagnostics into the diagnostic tool as soon as possible, those provides substantial information for Solutions troubleshooting (i.e. Elastic Security).

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+    */
+    public int runBasicQueries(RestClient client, DiagnosticContext context) {
+
+        int totalRetries = 0;
+        List<RestEntry> queries = new ArrayList<>();
+        
+        for (Map.Entry<String, RestEntry> entry : context.elasticRestCalls.entrySet()) {
+
+            String actionName = entry.getValue().getName().toString();
+            if (actionName.equals("kibana_alerts") || actionName.equals("kibana_detection_engine_find")) {
```
pickypg, line 132: Manually listing, in code, the requests that require paging is probably going to be annoying in the long term, but it's fine for now.

In the future, we should probably augment the YAML file that defines these APIs to allow it to specify whether it pages. (No action needed for today)

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+    /**
+    * On this function we will use the RestEntry action object to get the URL and execute the API one first time
+    * to get the total of events defined in Kibana for that API.
+    * Once we have the total of event we can calculate the number of pages/call we will need to execute
+    * then we call getNewEntryPage to create a new RestEntry for each page.
+    *
+    * @param  RestClient client
+    * @param  List<RestEntry> queries
+    * @param  double perPage - This is the number of docusment we reques to the API (set to 100 in execute func / or n in unit test)
+    * @param  RestEntry action
```
pickypg, line 153: Cleanup.

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+    *
+    * @return         void
```
pickypg, line 155: `void` functions do not return.

```suggestion
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+    */
+    public void getAllPages(RestClient client, List<RestEntry> queries, double perPage, RestEntry action) {
+        // get the values needed to the pagination.
+        RestResult res = client.execQuery(String.format("%s?per_page=1", action.getUrl()));
+        if (! res.isValid()) {
+            throw new DiagnosticException( res.formatStatusMessage( "Could not retrieve Kibana API pagination - unable to continue."));
+        }
+        String result   = res.toString();
+        JsonNode root   = JsonYamlUtils.createJsonNodeFromString(result);
+        double total    = root.path("total").intValue();
```
pickypg, line 165: Why is this a `double`? It's explicitly an `int` from how it's parsed.

```suggestion
        int total       = root.path("total").intValue();
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+        if (total > 0 && perPage > 0) {
+            // Get the first actions page
+            queries.add(getNewEntryPage(perPage, 1, action));
+            // If there is more pages add the new queries
+            if (perPage < total) {
+                double numberPages = Math.ceil(total/perPage);
```
pickypg, line 171: Seems like `double` trouble

```suggestion
                int numberPages = (int)Math.ceil(total / perPage);
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+
+    /**
+    * create a new RestEntry page for the original RestEntry action 
+    *
+    * @param  int perPage
+    * @param  int page
+    * @param  RestEntry action
+    * @return RestEntry
+    */
+    private RestEntry getNewEntryPage(double perPage, int page, RestEntry action) {
```
pickypg, line 187:
```suggestion
    private RestEntry getNewEntryPage(int perPage, int page, RestEntry action) {
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+                double numberPages = Math.ceil(total/perPage);
+                for (int i = 2; i <= numberPages; i++) {
+                    queries.add(getNewEntryPage(perPage, i, action));
+                }
+            }
+        }
+    }
+
+    /**
+    * create a new RestEntry page for the original RestEntry action
```
pickypg, line 180: This comment is basically saying "Wrap `RestEntry`" which tells us nothing.

```suggestion
    * Get a new `RestEntry` with the proper querystring parameters for paging.
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+    }
+
+    /**
+    * create a new RestEntry page for the original RestEntry action 
+    *
+    * @param  int perPage
+    * @param  int page
+    * @param  RestEntry action
+    * @return RestEntry
```
pickypg, line 185: Cleanup.

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+    * Extract the information on the kibana_node_stats.json with getNodeProfile function.
+    * Within the function getRemoteSystem or getLocalSystem we set ResourceCache.addSystemCommand
+    * Return will be used as workaround to Unit test. Will be replaced in v9
+    *
+    * @param  DiagnosticContext context
+    * @return SystemCommand
+    */
+    public SystemCommand execSystemCommands(DiagnosticContext context) {
+
+        ProcessProfile nodeProfile = getNodeProfile(context.tempDir, "kibana_node_stats.json", context);
```
pickypg, line 203:
```suggestion
        ProcessProfile profile = getProfile(context.tempDir, "kibana_stats.json", context);
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+
+
+    /**
+    * This function is executed **after** runBasicQueries
+    * Extract the information on the kibana_node_stats.json with getNodeProfile function.
+    * Within the function getRemoteSystem or getLocalSystem we set ResourceCache.addSystemCommand
+    * Return will be used as workaround to Unit test. Will be replaced in v9
+    *
+    * @param  DiagnosticContext context
+    * @return SystemCommand
```
pickypg, line 199: Cleanup.

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+    private RestEntry getNewEntryPage(double perPage, int page, RestEntry action) {
+        return new RestEntry(String.format("%s_%s", action.getName(), page), "", ".json", false, String.format("%s?per_page=%s&page=%s", action.getUrl(), perPage, page), false);
+    }
+
+
+    /**
+    * This function is executed **after** runBasicQueries
+    * Extract the information on the kibana_node_stats.json with getNodeProfile function.
+    * Within the function getRemoteSystem or getLocalSystem we set ResourceCache.addSystemCommand
+    * Return will be used as workaround to Unit test. Will be replaced in v9
```
pickypg, line 196: This is an ideal function. It takes parameters and, without any direct side effects, returns results. (There are some indirect side effects when the inner functions are invoked and the update the cache, which should be done by the caller, not _each_ of them)

```suggestion
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+                        }
+                    }
+                }
+            }
+            if (headerRemoved == true) {
+                String fileName = context.tempDir + SystemProperties.fileSeparator + "kibana_actions.json";
+                try (FileWriter fileWriter = new FileWriter(fileName)) {
+                    fileWriter.write(actions.toPrettyString());
+                    fileWriter.flush();
+                    fileWriter.close();
```
pickypg, line 271: `try`-with-resources automatically closes them.

```suggestion
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+        return syscmd;
+    }
+
+
+    /**
+    * We will not collect all the headerst hat are returned by the actions API (kibana_actions.json file).
+    * for troubleshooting support engineers will only need "kbn-xsrf" or "Content-Type", all the others are removed.
+    *
+    * @param  DiagnosticContext context
+    * @return void
```
pickypg, line 241: Cleanup.

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+    */
+    public void allowedHeadersFilter(DiagnosticContext context) {
```
pickypg, line 243: The name of this function is confusing. It sounds generic, but it's not generic at all.

Should this header removal be applied to all Kibana responses? Is headers a special part of the `kibana_actions` API response (and not the literal HTTP headers since the referenced headers _are_ HTTP headers that normally come back from every Kibana API)?

The function should be renamed (probably `filterActionsHeaders` based on reading it) and the answer to those questions should be at least inferable from function comment.

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+                // API webhook format can change, and maybe we have a webhook without config
+                if (!(config == null || config.isNull())) {
+                    // Not all the webhook need to have headers, so we need to be sure the data was set by the customer.
+                    JsonNode headers = actions.get(i).get("config").get("headers");
+                    if (!(headers == null || headers.isNull())) {
+                        Iterator<Map.Entry<String, JsonNode>> iter = actions.get(i).get("config").get("headers").fields();
+
+                        while (iter.hasNext()) {
+                            Map.Entry<String, JsonNode> entry = iter.next();
+                            if (!entry.getKey().equals("kbn-xsrf") && !entry.getKey().equals("Content-Type")) {
```
pickypg, line 258: It's worth noting that HTTP header keys are technically case insensitive. Kibana is almost certainly doing this consistently, but it's probably better to get the lowercased version of the key.

```suggestion
                            String key = entry.getKey().toLowerCase();

                            if (!key.equals("kbn-xsrf") && !key.equals("content-type")) {
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+                }
+            }
+            if (headerRemoved == true) {
+                String fileName = context.tempDir + SystemProperties.fileSeparator + "kibana_actions.json";
+                try (FileWriter fileWriter = new FileWriter(fileName)) {
+                    fileWriter.write(actions.toPrettyString());
+                    fileWriter.flush();
+                    fileWriter.close();
+                } catch (IOException e) {
+                  new RuntimeException("Error message", e).printStackTrace();
```
pickypg, line 273:
```suggestion
                  logger.error("Unexpected error while writing [kibana_actions.json]", e);
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+                            Map.Entry<String, JsonNode> entry = iter.next();
+                            if (!entry.getKey().equals("kbn-xsrf") && !entry.getKey().equals("Content-Type")) {
+                                iter.remove();
+                                headerRemoved = true;
+                            }
+                        }
+                    }
+                }
+            }
+            if (headerRemoved == true) {
```
pickypg, line 266: It's worth adding a comment here to explain why it's only needed if done.

```suggestion

            // If any headers were removed, we need to rewrite the file to remove them
            if (headerRemoved == true) {
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+
+
+    /**
+    * One of the requirements was test this new Kibana code.
+    * I splitted the "execute" function as in other examples (RunLogstashQueries) uses many static calls
+    * and create many new objects, make it difficult to mock and test.
+    * To be able to test the code without changing the global structure I splitted in smallest functions.
+    * Most of the new functions are public so I'm able to test it (Private if the test need to be done in a different class).
+    * When we will have more Unit test, the test for some functions here will be redundant and the public functions workaround will not needed anymore, 
+    * e.g. createJsonNodeFromFileName, this can be tested on the JsonYamlUtilsTest file (not existing in v8.1.2).
```
pickypg, line 287: Talk about what it does, not why there's a bunch of functions around it.

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+    * 
+    *
+    * @param  DiagnosticContext context
+    * @return         JavaPlatform object
```
pickypg, line 291: Cleanup.

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+        try {
+            context.perPage         = 100;
+            RestClient client       = ResourceCache.getRestClient(Constants.restInputHost);
+            int totalRetries        = runBasicQueries(client, context);
+            allowedHeadersFilter(context);
+            execSystemCommands(context);
+
+        } catch (Throwable t) {
+            logger.error( "Kibana Query error:", t);
+            throw new DiagnosticException(String.format("Error obtaining Kibana output and/or process id - will bypass the rest of processing.. %s", Constants.CHECK_LOG));
```
pickypg, line 304:
```suggestion
            throw new DiagnosticException(String.format("Error obtaining Kibana output and/or process id - will bypass the rest of processing. %s", Constants.CHECK_LOG));
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+
+        try {
+            context.perPage         = 100;
+            RestClient client       = ResourceCache.getRestClient(Constants.restInputHost);
+            int totalRetries        = runBasicQueries(client, context);
+            allowedHeadersFilter(context);
+            execSystemCommands(context);
+
+        } catch (Throwable t) {
+            logger.error( "Kibana Query error:", t);
```
pickypg, line 303: It's unlikely that you want throwable. The only reasonably expected `Throwable` is `OutOfMemoryError`, which you will not be in a state to recover from given that `context` is manipulated as it goes due to the overall design of the existing diagnostic.

```suggestion
        } catch (Exception e) {
            logger.error("Kibana Query error:", e);
```

src/main/java/com/elastic/support/util/LocalSystem.java
```diff
@@ -86,6 +86,39 @@ public void copyLogs(List<String> entries, String logDir, String targetDir) {
...
     }
 
+    /**
+    * On this function we will try to collect the journalctl logs, 
+    * Some services as Kibana installed with the RPM package will give the access to the logs using the journalctl command
+    *
+    * @param  String serviceName
+    * @param  String targetDir
+    *
+    * @return         void
```
pickypg, line 96: Cleanup.

src/main/java/com/elastic/support/util/RemoteSystem.java
```diff
@@ -185,6 +185,51 @@ public void copyLogs(List<String> entries, String logDir, String targetDir) {
...
 
+    /**
+    * On this function we will try to collect the journalctl logs, 
+    * Some services as Kibana installed with the RPM package will give the access to the logs using the journalctl command
+    * We will create a temperory file to copy those logs, then move them to the calling host.
+    *
+    * @param  String serviceName
+    * @param  String targetDir
+    *
+    * @return         void
```
pickypg, line 196: Cleanup.

src/main/java/com/elastic/support/util/RemoteSystem.java
```diff
@@ -185,6 +185,51 @@ public void copyLogs(List<String> entries, String logDir, String targetDir) {
...
+    */
+    @Override
+    public void copyLogsFromJournalctl(String serviceName, String targetDir) {
+
+        ChannelSftp channelSftp = null;
+        String tempDir = "templogs";
+        String mkdir = "mkdir templogs";
+        String journalctl = " journalctl -u {{SERVICE}} > '{{TEMP}}/{{SERVICE}}.log'";
```
pickypg, line 204: It's odd seeing structural differences between clearly copy/pasted code (`LocalSystem:: copyLogsFromJournalctl`). These string replacements won't trigger errors, unless the parameters are not trustworthy and are for-sure non-`null`.

src/main/java/com/elastic/support/util/SystemUtils.java
```diff
@@ -162,11 +162,21 @@ public static String getCurrentTime() {
...
 
+    /**
+    * This function is key for two types or execution remote and local.
+    * to collect some system data or logs we need to be able to know the OS where the process is running
+    * Logstash is not using this as we only collect the API data, no system data collected on the servers.
+    * "win32" is used by Kibana API as OS name.
+    *
+    * @param  String osName , this value is provided by the Elasticsearch nodes or Kibana stats API
+    *
+    * @return         String
```
pickypg, line 173: Cleanup.

pickypg (changes requested):
Some of this is nitpicks, some of it's structural.

src/main/java/com/elastic/support/diagnostics/DiagnosticInputs.java
```diff
@@ -89,7 +101,7 @@
 
     public final static String  typeDescription = "Enter the number of the diagnostic type to run.";
     public final static String  remoteUserDescription = "User account to be used for running system commands and obtaining logs. This account must have sufficient authority to run the commands and access the logs.";
-    public final static String  remotePasswordDescription = "Password for the remote login.";
+    public final static String  remotePasswordDescription = "Password for the remote login:";
```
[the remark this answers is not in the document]
CamiloSierraH: This come from an issue we had when testing the diagnostic with Ed. In this moment the diagnostic is requesting to the user to add the remote login, when we were running this, as this was a dot, we don't understand why the diagnostic stop there.
We though that the diagnostic get stuck and that was the last log displayed on the screen.
The colon will expect that will help the user to understand that we are asking for the password.

src/main/java/com/elastic/support/diagnostics/commands/CheckKibanaVersion.java
```diff
@@ -0,0 +1,107 @@
...
+    public void execute(DiagnosticContext context) {
+
+        // Get the version number from the JSON returned
+        // by just submitting the host/port combo
+        logger.info(Constants.CONSOLE, "Getting Kibana Version.");
+        DiagnosticInputs inputs = context.diagnosticInputs;
+
+        try {
+           RestClient restClient = RestClient.getClient(
+                    context.diagnosticInputs.host,
```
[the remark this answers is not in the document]
CamiloSierraH: I know, when doing the unit-test add all the variables on the order was a nightmare. If I'm not wrong this was changed in v9.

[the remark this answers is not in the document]
pickypg: Perhaps, but I don't see us moving to the existing v9 code.

src/main/java/com/elastic/support/diagnostics/commands/CollectSystemCalls.java
```diff
@@ -38,7 +38,7 @@ public void execute(DiagnosticContext context) {
             // Get the configurations for that platoform's sys calls.
             Map<String, String> sysCalls = osCmds.get("sys");
             processCalls(targetDir, sysCalls, sysCmd, pid);
-
+            logger.info( Constants.CONSOLE, "First set of system calls executed.");
```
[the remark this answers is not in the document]
CamiloSierraH: I added this log because this file do two different kind of request, and most of the time the second request fails and in the logs we only see `Unexpected error - bypassing some or all system calls.`.

The first group of request are the `sysstat` 99% of the time they works, but checking the logs we see only the error on the logs.
The request that fail are `javac`, `jps`... and this produce the error on the logs.

The try/catch cover the whole function :/ so this log it was added to add more info in the diagnostic output log file, so we know that at least the first set of system calls were executed.

I thought that this information was interesting, but I can remove it... and in the future we can split the function so we can have more specific logs and errors.

src/main/java/com/elastic/support/diagnostics/DiagnosticInputs.java
```diff
@@ -241,13 +253,24 @@ public boolean runInteractive() {
...
         if (val.toLowerCase().contains("logstash")) {
             if (port == 9200) {
                 port = Constants.LOGSTASH_PORT;
             }
+        } else if (val.toLowerCase().contains("kibana")) {
+            if (port == 9200) {
+                port = Constants.KIBANA_PORT;
+            }
         }
         return null;
```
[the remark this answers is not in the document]
CamiloSierraH: here some spaghetti, so the parent function is expecting null as response, so they will use the default. but the default is set/updated within the function that returns null. (the port we change is a public variable defined on the parent class ElasticRestClientInputs)
In a few words if we do not return null the default is not applied/updated, and we think there is an error/problem with the port defined.

But I'm not finished yet, the function is also defined in BaseInputs (parent of ElasticRestClientInputs that its parent of DiagnosticInputs) 
In BaseInputs the function really return a list, and null if the customer has set a correct port, the list returned was in case the port defined was incorrect (val < 1 || val > 65535), So it make sense in BaseInputs, but not anymore in DiagnosticInputs (as we always return null).

I was checking and all the other validate functions on the DiagnosticInputs class worked the same.
I'm not sure if the main idea was some kind of Factory design and then by mistake we add global variables...

src/main/java/com/elastic/support/diagnostics/commands/CheckKibanaVersion.java
```diff
@@ -0,0 +1,107 @@
...
+
+public class CheckKibanaVersion implements Command {
+
+    /**
+     * Gets the version of Kibana that is running. This also
+     * acts as a sanity check. If there are connection issues and it fails
+     * this will bet the first indication since this is lightweight enough
+     * that is should usually succeed. If we don't have a version we
+     * won't be able to generate the correct call selection later on.
+     */
```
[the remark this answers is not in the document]
CamiloSierraH: sorry this was the doc for the class itself, I will move it to the class level.

src/main/java/com/elastic/support/diagnostics/commands/CheckKibanaVersion.java
```diff
@@ -0,0 +1,107 @@
...
+            RestResult res = client.execQuery("/api/settings");
+            if (! res.isValid()) {
+                throw new DiagnosticException( res.formatStatusMessage( "Could not retrieve the Kibana version - unable to continue."));
+            }
+            String result = res.toString();
+            JsonNode root = JsonYamlUtils.createJsonNodeFromString(result);
+            String version = root.path("settings").path("kibana").path("version").asText();
+            logger.info(Constants.CONSOLE, String.format("Kibana Version is :%s", version));
+
+            Pattern versionPattern = Pattern.compile("([1-9]\\d*)\\.(\\d+)\\.(\\d+)?");
```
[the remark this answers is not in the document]
CamiloSierraH: I did the change and add one more unit test (we already had 3 other test checking correct, empty or wrong versions)

src/main/resources/kibana-rest.yml
```diff
@@ -0,0 +1,47 @@
...
+#   NPM Versioning rules are documented here: https://github.com/npm/node-semver
+#
+#   Note to those adding entries: within each group, cat API's, json API's, and commercial, they are in alphabetical order.
+#   Please adhere to this convention when submitting pull requests.
+
+kibana_settings:
+  versions:
+    ">= 6.5.0": "/api/settings"
+
+kibana_node_stats:
```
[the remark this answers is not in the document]
CamiloSierraH: I updated the name and the unit test!!

src/main/java/com/elastic/support/diagnostics/commands/KibanaCheckPlatformDetails.java
```diff
@@ -0,0 +1,193 @@
...
+            List<ProcessProfile> nodeProfiles = getNodeNetworkAndLogInfo(infoNodes);
+
+            // See if this cluster is dockerized - if so, don't bother checking for a master to go to
+            // because the port information in its output is not reliable.
+            for (ProcessProfile profile : nodeProfiles) {
+                if (profile.isDocker) {
+                    context.dockerPresent = true;
+                    break;
+                }
+            }
```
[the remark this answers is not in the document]
CamiloSierraH: I add a new private function for the is running in docker code, but I keep on the main function:
```
List<ProcessProfile> profiles = getNodeNetworkAndLogInfo(infoProcess);
```
BTW I replace `node` by process or profiles!! As mentioned before the term node was misleading in Kibana.

src/main/java/com/elastic/support/diagnostics/commands/KibanaCheckPlatformDetails.java
```diff
@@ -0,0 +1,193 @@
...
+            RestClient restClient = ResourceCache.getRestClient(Constants.restInputHost);
+
+            // Populate the node metadata
+            Map<String, RestEntry> calls = context.elasticRestCalls;
+            RestEntry entry = calls.get("kibana_node_stats");
+            RestResult result = restClient.execQuery(entry.getUrl());
+
+            // Initialize to empty node which mimimizes NPE opportunities
+            JsonNode infoNodes = JsonYamlUtils.mapper.createObjectNode();
+            if (result.getStatus() == 200) {
```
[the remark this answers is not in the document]
CamiloSierraH: Done, I also added a new unit test with the status != 200, I had only tests with HTTP status 200.

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
CamiloSierraH: I did this change in a different commit!

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,309 @@
...
+        return syscmd;
+    }
+
+    /**
+    * On this function we will get create an LocalSystem object
+    * depending the parseOperatingSystem value it will override or not the value of osName with the constant SystemProperties.osName
+    *
+    * @param  String osName
+    * @param  Boolean parseOperatingSystem
+    * @return         LocalSystem object
```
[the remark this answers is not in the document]
CamiloSierraH: As we already have the local or remote defined I remove this function!

src/main/java/com/elastic/support/diagnostics/commands/KibanaCheckPlatformDetails.java
```diff
@@ -0,0 +1,193 @@
...
+            if (diagNode.pid.equals("1")) {
+                diagNode.isDocker = true;
+            }
+
+            diagNode.networkHost = nodesInfo.path("kibana").path("host").asText();
+            diagNode.host = nodesInfo.path("kibana").path("host").asText();
+
+            String nodeOs = nodesInfo.path("os").path("platform").asText().toLowerCase();
+            diagNode.os = SystemUtils.parseOperatingSystemName(nodeOs);
+            diagNode.javaPlatform = new JavaPlatform(diagNode.os);
```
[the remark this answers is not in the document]
CamiloSierraH: It does not, but we use this to access later to the diag.yml values... So I remove it, but for that I adapt the code in CollectSystemCalls.

src/main/java/com/elastic/support/diagnostics/DiagnosticInputs.java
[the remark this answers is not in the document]
CamiloSierraH: BTW, if you run the diagnostic, before this question the OS will ask you the sudo password like this:
```
[sudo] password for camilo: 
```
So it's weird, the first one use colon and then we ask with a dot:
```
Password for the remote login.
```
For me it will make more sense if we follow the OS and when asking a question we use colon.

README.md
```diff
@@ -1,108 +1,108 @@
...
+     - [Monitoring Data Import Options](#monitoring-data-import-options)
+     - [Monitoring Import Examples](#monitoring-import-examples)
+   - [Standard Diagnostic Troubleshooting](#standard-diagnostic-troubleshooting)
+ 
+ # Support Diagnostics Utility
+ 
+ [Click here for the latest version of the Support Diagnostics Utility](https://github.com/elastic/support-diagnostics/releases/latest)
+ 
+ The support diagnostic utility is a Java application that can interrogate a running Elasticsearch cluster, Kibana or Logstash process to obtain data about the state of the cluster at that point in time. It is compatible with all versions of Elasticsearch (including alpha, beta and release candidates), and for Logstash versions greater than v5, and for Kibana greater than v6.5.  The release version of the diagnostic is independent of the Elasticsearch, Kibana or Logstash version it is being run against. If it cannot match the targeted version it will attempt to run calls from the latest configured release. Linux, OSX, or Windows platforms are all supported, and it can be run as a standalone utility or from within a Docker container.
```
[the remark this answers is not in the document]
CamiloSierraH: I have not added this text for logstash, but checking the `logstash-rest.yml` : 
```
logstash_version:
  versions:
    "> 0.0.0": "/?pretty"

logstash_node:
  versions:
    "> 0.0.0": "/_node?pretty"

logstash_node_stats:
  versions:
    "> 0.0.0": "/_node/stats?pretty"

logstash_nodes_hot_threads:
  versions:
    "> 0.0.0": "/_node/hot_threads?human=true&threads=10000"

logstash_plugins:
  versions:
    "> 0.0.0": "/_node/plugins?pretty"
```
So I guess that any Logstash version will be executed... but on the documentation of Logstash I can only see the API since v5.5.

Anyway I will take the README in v8.3, and add my changes!

src/main/java/com/elastic/support/diagnostics/commands/CollectSystemCalls.java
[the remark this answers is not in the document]
pickypg: That's reasonable. Let's just remove that random leading space then:

```suggestion
            logger.info(Constants.CONSOLE, "First set of system calls executed.");
```

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+ * If this request fails, then the rest of the diagnostic cannot proceed, as we need to 
+ * have enough details to connect to Kibana instace.
+ */
+public class KibanaGetDetails extends CheckPlatformDetails {
+
+    /**
+    * Collect server details from Kibana using the configured {@code context}, and store
+    * updated details back into the {@code context}.
+    * @param  context
+    */
```
pickypg, line 41:
```suggestion
     * Collect server details from Kibana using the configured {@code context}, and store
     * updated details back into the {@code context}.
     * @param context The diagnostic context
     */
```

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+
+     /**
+    * If Kibana process is running on a remote server 
+    * we need to check if is a docker process or if is running as any other process
+    * Then create a remote system to request the Kibana process
+    *
+    * @param  context
+    * @param  profiles
```
pickypg, line 79: Add descriptions

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+    */
```
pickypg, line 80:
```suggestion
    /**
     * If Kibana process is running on a remote server 
     * we need to check if is a docker process or if is running as any other process
     * Then create a remote system to request the Kibana process
     *
     * @param  context
     * @param  profiles
     */
```

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+
+    /**
+    * If Kibana process is running on the same server 
+    * add a loca system command to be able to execute the API requests
+    *
+    * @param  context
+    * @param  profiles
+    */
```
pickypg, line 120:
```suggestion
     * If Kibana process is running on the same server 
     * add a loca system command to be able to execute the API requests
     *
     * @param  context
     * @param  profiles
     */
```

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+        ));
+    }
+
+
+    /**
+    * If Kibana process is running on the same server 
+    * add a loca system command to be able to execute the API requests
+    *
+    * @param  context
+    * @param  profiles
```
pickypg, line 119: Add descriptions

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+
+    /**
+    * Map Kibana / stats API results to the ProcessProfile object.
+    *
+    * @param  processInfo all the context sent by the kibana stats API
+    *
+    * @return The network info as defined in the kibana stats API
+    */
```
pickypg, line 151:
```suggestion
     * Map Kibana / stats API results to the ProcessProfile object.
     *
     * @param  processInfo all the context sent by the kibana stats API
     * @return The network info as defined in the kibana stats API
     */
```

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+    public List<ProcessProfile> getNodeNetworkAndLogInfo(JsonNode processInfo) {
+
+        List<ProcessProfile> nodeNetworkInfo = new ArrayList<>();
```
pickypg, line 154:
```suggestion
        List<ProcessProfile> nodeNetworkInfo = new ArrayList<>();

```

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+
+    /**
+    * Get the Kibana server instance's profile.
+    *
+    * @param  host
+    * @param  profiles list of network information for each kibana instance running
+    *
+    * @return return the profile for the kibana process
+    * @throws RuntimeException if there is not exactly one profile.
+    */
```
pickypg, line 190:
```suggestion
     * Get the Kibana server instance's profile.
     *
     * @param  host
     * @param  profiles list of network information for each kibana instance running
     * @return return the profile for the kibana process
     * @throws RuntimeException if there is not exactly one profile.
     */
```

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+        return  profiles.get(0);
+    }
+
+    /**
+     * Fetch the Kibana Stats JSON payload from the Kibana server.
+     *
+     * @param context The current diagnostic context
+     * @return Never {@code null}.
+     * @throws DiagnosticContext if Kibana responds with a non-200 response
+    */
```
pickypg, line 226:
```suggestion
     */
```

src/main/java/com/elastic/support/diagnostics/commands/KibanaGetDetails.java
```diff
@@ -0,0 +1,245 @@
...
+
+    /**
+    * Get the Kibana server instance's profile.
+    *
+    * @param  host
+    * @param  profiles list of network information for each kibana instance running
+    *
+    * @return return the profile for the kibana process
+    * @throws RuntimeException if there is not exactly one profile.
+    */
```
pickypg, line 210:
```suggestion
     * Get the Kibana server instance's profile.
     *
     * @param profiles list of network information for each kibana instance running
     * @return The first profile.
     * @throws RuntimeException if there is not exactly one profile.
     */
```

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,256 @@
...
+
+/**
+ * RunKibanaQueries executes the version-dependent REST API calls against Kibana.
+ */
+public class RunKibanaQueries extends BaseQuery {
+
+    private static final Logger logger = LogManager.getLogger(RunKibanaQueries.class);
+
+    /**
+    * Create a new ProcessProfile object and extract the information from fileName to get PID and OS.
```
pickypg, line 35: Line these up consistently please (don't want to just keep filling up the review).

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,256 @@
...
+    * @param  tempDir
+    * @param  fileName
+    * @param  context
```
pickypg, line 38: Add descriptions

src/main/java/com/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+
+    /**
+    * CheckKibanaVersion (executed before) defined/set the context.elasticRestCalls.
+    * here we will loop and create a List of RestEntry with the queries/APIs that need to be executed
+    * You have two APIs that work differently and and may have or not many pages, so we use the getAllPages function
+    * runQueries called in this function create a json file for each RestEntry set on the 'queries' variable.
+    *
+    * @param  client
```
pickypg, line 58: Please be sure to add appropriate descriptions.

src/main/java/com/elastic/support/util/LocalSystem.java
```diff
@@ -86,6 +86,38 @@ public void copyLogs(List<String> entries, String logDir, String targetDir) {
...
+    *
+    * @param  serviceName
+    * @param  targetDir
+    */
+    @Override
+    public void copyLogsFromJournalctl(String serviceName, String targetDir) {
+
+        String tempDir = "templogs";
+        String mkdir = "mkdir templogs";
+        String journalctl = " journalctl -u {{SERVICE}} > '{{TEMP}}/{{SERVICE}}.log'";
```
pickypg, line 101: Does this need a leading space for the `runCommand`?

```suggestion
        String journalctl = "journalctl -u {{SERVICE}} > '{{TEMP}}/{{SERVICE}}.log'";
```

src/main/java/com/elastic/support/util/LocalSystem.java
```diff
@@ -86,6 +86,38 @@ public void copyLogs(List<String> entries, String logDir, String targetDir) {
...
+        try {
+            runCommand(mkdir);
+            runCommand(journalctl);
+            String source = "{{TEMP}}/{{SERVICE}}.log";
+            source = source.replace("{{SERVICE}}", serviceName);
+            source = source.replace("{{TEMP}}", tempDir);
+            String target = targetDir + SystemProperties.fileSeparator + serviceName;
+            FileUtils.copyFile(new File(source), new File(target));
+            // clean up the temp logs on the remote host
+            runCommand(" rm -Rf templogs");
```
pickypg, line 114: Same comment:

```suggestion
            runCommand("rm -Rf templogs");
```

src/main/java/com/elastic/support/util/RemoteSystem.java
```diff
@@ -185,6 +185,48 @@ public void copyLogs(List<String> entries, String logDir, String targetDir) {
         }
     }
 
+    /**
+    * Some services as Kibana installed with the RPM package will give the access to the logs using the journalctl command
+    * We will create a temperory file to copy those logs, then move them to the calling host.
+    *
+    * @param  serviceName
```
pickypg, line 192: Just a note this is in other files like this.

src/main/java/com/elastic/support/util/RemoteSystem.java
```diff
@@ -185,6 +185,48 @@ public void copyLogs(List<String> entries, String logDir, String targetDir) {
...
+    * @param  targetDir
+    */
+    @Override
+    public void copyLogsFromJournalctl(String serviceName, String targetDir) {
+
+        ChannelSftp channelSftp = null;
+        String tempDir = "templogs";
+        String mkdir = "mkdir templogs";
+        String journalctl = " journalctl -u {{SERVICE}} > '{{TEMP}}/{{SERVICE}}.log'";
```
pickypg, line 201: Same comment:

```suggestion
        String journalctl = "journalctl -u {{SERVICE}} > '{{TEMP}}/{{SERVICE}}.log'";
```

src/main/java/com/elastic/support/util/RemoteSystem.java
```diff
@@ -185,6 +185,48 @@ public void copyLogs(List<String> entries, String logDir, String targetDir) {
...
+            // Now they have been moved from the log directory into the temp dir created in the user's home
+            // we can move them to the the calling host. It should pull them all in one pass.
+            channelSftp = (ChannelSftp) session.openChannel("sftp");
+            channelSftp.connect();
+            channelSftp.get("templogs/*", targetDir);
+            channelSftp.exit();
+
+            // clean up the temp logs on the remote host
+            runCommand(" rm -Rf templogs");
+
```
pickypg, line 218:
```suggestion
            runCommand("rm -Rf templogs");
```

src/main/resources/kibana-rest.yml
```diff
@@ -0,0 +1,47 @@
...
+  versions:
+    "> 7.10.0": "/api/detection_engine/prepackaged"
+
+kibana_alerts:
+  versions:
+    ">= 7.9.0": "/api/alerts/_find"
+
+kibana_task_manager_help:
+  versions:
+    ">= 7.11.0": "/api/task_manager/_health"
```
pickypg, line 47: Please sort these alphabetically. That will make it easier to find them later as this list grows.

pickypg (commented):
One last batch of nitpicks.

src/main/resources/kibana-rest.yml
CamiloSierraH: in the other files elastic and logstash this list is order by version, so you start with v0.9 and the at the end the latest API, I try to follow just the other two rest files. if we order this one by alpha and the other two by version it will be even harder to understand.

README.md
```diff
@@ -388,6 +405,18 @@ Executing Logstash diagnostics with a non-default port
 sudo ./diagnostics.sh --host 10.0.0.20 --type logstash-local --port 9607
 ```
 
+Executing Kibana diagnostics in the same server where kibana is running
+ 
+   ```$xslt
+   sudo ./diagnostics.sh --host localhost --port 5601 --type kibana-local
+   ```
```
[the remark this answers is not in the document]
CamiloSierraH: all the others ES, Logstash in the README has the same format.

README.md
```diff
@@ -388,6 +405,18 @@ Executing Logstash diagnostics with a non-default port
...
+
+Running the kibana-api type to suppress system call and log collection and explicitly configuring an output directory (this is also the option that need bo used when collecting the diagnotic for Kibana in **Elastic cloud**).
+ 
+   ```$xslt
```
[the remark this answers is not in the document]
CamiloSierraH: all the others ES, Logstash in the README has the same format.

src/main/resources/kibana-rest.yml
pickypg: I strongly disagree. The other files were sorted by what is, in effect, a random order. If this were a table, it would make sense, but you shouldn't be diving through a file looking for content by an inner field.

The impact will be that Kibana will be different, but the moment Logstash or ES are changed, I'll have those files sorted.

README.md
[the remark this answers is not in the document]
pickypg: To be explicit, no: they're not indented.

README.md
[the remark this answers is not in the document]
pickypg: To be a bit more explicit, here is the text copied from the raw file with your changes (I escapted the \` marks to avoid it terminating the code block). Notice the indentation difference:

```
Executing Logstash diagnostics with a non-default port

\`\`\`$xslt
sudo ./diagnostics.sh --host 10.0.0.20 --type logstash-local --port 9607
\`\`\`

Executing Kibana diagnostics locally from the same server where Kibana is running
 
   \`\`\`$xslt
   sudo ./diagnostics.sh --host localhost --port 5601 --type kibana-local
   \`\`\`

Running the `kibana-api` type to suppress system call and log collection and explicitly configuring an output directory (this is also the option that needs to be used when collecting the diagnostic for Kibana in **Elastic Cloud**).
 
   \`\`\`$xslt
   sudo ./diagnostics.sh --host 2775abprd8230d55d11e5edc86752260dd.us-east-1.aws.found.io --port 9243 --type kibana-api -u elastic --password --ssl -o /home/user1/diag-out
   \`\`\`
 
Executing against a remote host with full collection, using sudo, and enabling trust where there's no known host entry. Note that the diagnostic is not executed via sudo because all the privileged access is on a different host.

\`\`\`$xslt
./diagnostics.sh --host 10.0.0.20 --type remote -u someuser --password --ssl --remoteUser someuser --remotePass --trustRemote --sudo`
\`\`\`
```

src/main/resources/kibana-rest.yml
CamiloSierraH: great I will do the change!

pickypg (approved):
LGTM. I know you planned to re-run local tests again, so once you get back with those results we can merge this and do a release.

CamiloSierraH:
Hey @pickypg, thanks for your help, I just tested the Kibana type in Remote, API, and Local and they are working well!

Issue #485, opened 2021-04-26
Support Diagnostics Tool consumed as a Java library


brunofarache:
The Support Diagnostics Tool was designed to be used from the command line.

The objective of this issue is to also make it possible to consume this tool as a Java library.

Here are the changes to make that happen:

  - [x] Include `*.yaml` config files in released jars. Instead of distributing `*.yaml` separately from the jar file, it would be better to include them in the binary, this way, developers can just drop the jar in their projects without having to include a separate `config` folder in the classpath. (#488)
  - [x] `DiagnosticService#exec` should return `java.io.File` instead of `void`. This makes it easier for callers to know where the zip file was saved, otherwise they would need to list which files are present in the `outputDir` after the tool is finished. (#486)
  - [x] `DiagnosticService#exec` should throw `DiagnosticException` instead of logging and swallowing errors. `DiagnosticException` is a `RuntimeException` which makes harder for consumers be aware of errors and catch them properly. (#486)
  - [x] Close all `OutputStream` correctly to avoid memory leaks. I noticed some `OutputStream` instances are not being closed properly, while that's fine when running the tool in the command line, it may leak memory when being used many times by the same process. (#487)
  - [x] Archive should be always created, regardless if exceptions were thrown or not. (#493)
  - [x] Make sure the generated bundle contains `diagnostics.log`.
  - [x] Add option to always generate zip bundles and skip tars. (#494)
  - [x] Change license to ELv2. (#501)
  - [x] Publish to private Maven repo. We need to publish it somewhere in order to be added as a dependency to Java projects.

CBlaisure:
Grooming notes:
Support Development to investigate github actions to deploy Maven automatically.

shasts:
@brunofarache 
Just so you know, we have a few internal libraries published in our artifactory.

pickypg:
I was able to get the person with Maven Central access to give me approval to create packages in the `co.elastic.*` `groupId`. They were on vacation until today. Now we wait for Sonatype.

https://issues.sonatype.org/browse/OSSRH-70696

They literally gave me permissions as I was posting this. I'll be trying it out soon.

pickypg:
@brunofarache I was able to get this deployed to Maven Central: version 8.2.0. Gradle dependency:

```
dependencies {
  compile 'co.elastic.support:diagnostics:8.2.0'
}
```

pickypg:
I was able to confirm this works locally with my own cluster:

```java
[...]

PR #487, opened 2021-04-27
Close streams properly

brunofarache:
Part of #485.

In many places `Closeable` instances such as `OutputStream`, `InputStream` and `Writer` weren't being closed properly in finally blocks. This may cause memory leaks if exceptions are thrown while writing/reading to those streams.

Tried to use try-with-resources as much as possible, only in few cases closing in finally blocks made sense.

### DIFF
--- a/src/main/java/com/elastic/support/util/JsonYamlUtils.java
+++ b/src/main/java/com/elastic/support/util/JsonYamlUtils.java
@@ -3,9 +3,7 @@ package com.elastic.support.util;
 import com.elastic.support.diagnostics.DiagnosticException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


src/main/java/com/elastic/support/util/JsonYamlUtils.java
```diff
@@ -56,118 +46,69 @@ public static JsonNode createJsonNodeFromString(String nodeString) {
...
-      finally {
-         return nullSafeYamlMap(doc);
-      }
-   }
+      return nullSafeYamlMap(doc);
+}
 
-   public static Map flattenYaml(Map map) {
-      Map result = new LinkedHashMap<>();
+   public static Map<String, Object> flattenYaml(Map<String, Object> map) {
```
pickypg, line 74: Thanks for making these actually use generics properly.

src/main/java/com/elastic/support/util/JsonYamlUtils.java
```diff
@@ -181,21 +122,11 @@ public static void buildFlattenedMap(Map result, Map source, String path) {
...
-      }
-      return doc;
-   }
-
-   private static Map listToMap(List input){
-      int sz = input.size();
-      Map output = new LinkedHashMap<>();
-      for(int i=0; i < sz; i++){
-         output.put("idx_" + i, input.get(i));
+         doc = new HashMap<>();
```
pickypg, line 127: Nitpick: no reason to touch the reference.

```suggestion
         return new HashMap<>();
```

pickypg (approved):
LGTM with a nitpick.

src/main/java/com/elastic/support/util/JsonYamlUtils.java
brunofarache: Done and also removed some other unused code in https://github.com/elastic/support-diagnostics/pull/487/commits/279345dff9ef69ad6cbd481fae771f7f8a210743.

Issue #489, opened 2021-04-29
domain login fails with support-diagnostics


wolframhaussig:
I wanted to look into a problem on our elastic stack using the support diagnostics tool. As we primarily use the active directory login I tried to use the tool with my personal account which has superuser privileges on the elastic cluster. Unfortunately, I get the following error:

```
ERROR com.elastic.support.diagnostics.chain.DiagnosticChainExec - Unexpected error
java.lang.IllegalArgumentException: Illegal character in path at index 21: /_security/user/mydomain\myuser
	at java.net.URI.create(URI.java:883) ~[?:?]
	at org.apache.http.client.methods.HttpGet.<init>(HttpGet.java:66) ~[httpclient-4.5.10.jar:4.5.10]
	at com.elastic.support.rest.RestClient.execGet(RestClient.java:66) ~[support-diagnostics-8.1.4.jar:8.1.4]
	at com.elastic.support.rest.RestClient.execQuery(RestClient.java:58) ~[support-diagnostics-8.1.4.jar:8.1.4]
	at com.elastic.support.diagnostics.commands.CheckUserAuthLevel.execute(CheckUserAuthLevel.java:42) ~[support-diagnostics-8.1.4.jar:8.1.4]
	at com.elastic.support.diagnostics.chain.DiagnosticChainExec.runDiagnostic(DiagnosticChainExec.java:31) [support-diagnostics-8.1.4.jar:8.1.4]
	at com.elastic.support.diagnostics.DiagnosticService.exec(DiagnosticService.java:68) [support-diagnostics-8.1.4.jar:8.1.4]
	at com.elastic.support.diagnostics.DiagnosticApp.main(DiagnosticApp.java:42) [support-diagnostics-8.1.4.jar:8.1.4]
Caused by: java.net.URISyntaxException: Illegal character in path at index 21: /_security/user/mydomain\myuser
	at java.net.URI$Parser.fail(URI.java:2913) ~[?:?]
	at java.net.URI$Parser.checkChars(URI.java:3084) ~[?:?]
	at java.net.URI$Parser.parseHierarchical(URI.java:3166) ~[?:?]
	at java.net.URI$Parser.parse(URI.java:3125) ~[?:?]
	at java.net.URI.<init>(URI.java:600) ~[?:?]
	at java.net.URI.create(URI.java:881) ~[?:?]
	... 7 more
```

Of course I can use the default elastic superuser but my question is: Am I doing something wrong or is the support tool not designed for active directory?

[...]

PR #491, opened 2021-05-05
URL Encode Username

pickypg:
This URL encodes the username when supplying it as part of the URL
(rather than as part of an auth header).

I tested this by creating a user named `temp\user`. Using the existing diagnostic fails as noted in the linked issue. After using the code in this branch, it works successfully.

Closes https://github.com/elastic/support-diagnostics/issues/489


jpcarey (approved):
LGTM

PR #493, opened 2021-05-10
Should always create archive file when failures happen

brunofarache:
In #486, I changed the order in which `createArchive` was called, leaving it outside the `finally` block.

This may be problematic because the zip fule won't be created if `DiagnosticChainExec.runDiagnostic` fails with a `DiagnosticException`.

I'm changing it back to make it work as it used to work, that is, exceptions should always be printed to console AND to the `diagnostics.log` file inside the zip file.


brunofarache:
cc @pickypg

pickypg (approved):
LGTM

PR #494, opened 2021-05-10
Add archive type option

brunofarache:
Part of #485.

Right, the utility tries to zip the output dir, if zip fails, it fallbacks to tar.gz.

However, this should be deterministic, devs should be able to choose which output file type they want: zip or tar.gz.

### DIFF
--- a/src/main/java/com/elastic/support/BaseInputs.java
+++ b/src/main/java/com/elastic/support/BaseInputs.java
@@ -3,6 +3,7 @@ package com.elastic.support;
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.Parameter;
 import com.elastic.support.diagnostics.ShowHelpException;
+import com.elastic.support.util.ArchiveUtils.ArchiveType;
 import com.elastic.support.util.ResourceCache;
 import com.elastic.support.util.SystemProperties;
 
@@ -23,6 +24,7 @@ public abstract class BaseInputs {
     private static final Logger logger = LogManager.getLogger(BaseInputs.class);
     public static final String outputDirDescription = "Fully qualified path to an output directory. If it does not exist the diagnostic will attempt to create it. If not specified the diagnostic directory will be used: ";
     public static final String bypassDiagVerifyDescription = "Bypass the diagnostic version check. Use when internet outbound HTTP access is blocked by a firewall.";
+    public static final String archiveTypeDescription = "Output file type. Choose between: 'zip', 'tar' or 'any'. 'any' will try to zip first and fallback to tar if the zip fails. Defaults to any.";
     protected List<String> emptyList = new ArrayList<>();
     protected JCommander jCommander;
 
@@ -39,6 +41,9 @@ public abstract class BaseInputs {
     // Stop the diag from checking itself for latest version.
     @Parameter(names = {"--bypassDiagVerify"}, description = bypassDiagVerifyDescription)
     public boolean bypassDiagVerify = false;
+
+    @Parameter(names = {"--archiveType"}, description = archiveTypeDescription)
+    public String archiveType = "any";
 
     // End Input Fields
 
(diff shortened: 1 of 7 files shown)


brunofarache:
cc @pickypg

src/main/java/com/elastic/support/util/ArchiveUtils.java
```diff
@@ -18,24 +18,52 @@
...
+            try {
+               return createZipArchive(dir, archiveFileName);
+            } catch (IOException ioe) {
+               throw new DiagnosticException("Couldn't create zip archive.", ioe);
+            }
+         case TAR:
+            try {
+               return createTarArchive(dir, archiveFileName);
+            } catch (IOException ioe) {
+               throw new DiagnosticException("Couldn't create tar archive.", ioe);
```
pickypg, line 51:
```suggestion
               throw new DiagnosticException("Couldn't create tar.gz archive.", ioe);
```

src/main/java/com/elastic/support/monitoring/MonitoringExportService.java
```diff
@@ -4,10 +4,8 @@
 import com.elastic.support.diagnostics.DiagnosticException;
 import com.elastic.support.diagnostics.commands.CheckElasticsearchVersion;
 import com.elastic.support.rest.*;
-import com.elastic.support.util.JsonYamlUtils;
-import com.elastic.support.util.ResourceCache;
-import com.elastic.support.util.SystemProperties;
-import com.elastic.support.util.SystemUtils;
+import com.elastic.support.util.*;
```
pickypg, line 7: Not a fan of `*` imports, but it's fine for this since the code already has others.

src/main/java/com/elastic/support/monitoring/MonitoringExportInputs.java
```diff
@@ -132,8 +132,7 @@ public boolean runInteractive() {
             return Collections.singletonList("Invalid Date or Time format. Please enter the date in format YYYY-MM-dd HH:mm");
         }
 
-        return emptyList;
-
+        return null;
```
pickypg, line 135: Careful with returning `null` instead of an empty list. Some parts of the code expect things that they cannot safely expect.

Given that this code relates explicitly to fetching stack monitoring data, I am unsure if you tested this behavior.

src/main/java/com/elastic/support/BaseInputs.java
```diff
@@ -128,6 +145,16 @@ protected void runOutputDirInteractive(){
...
     }
 
+    public List<String> validateArchiveType(String archiveType) {
+        for (ArchiveType type : ArchiveType.values()) {
+            if (type.name().equalsIgnoreCase(archiveType)) {
+                return null;
+            }
+        }
+
+        return Collections.singletonList("Provided archive type" + archiveType + " is not supported.");
```
pickypg, line 155: Nitpick improvement for readability:

```suggestion
        return Collections.singletonList("Provided archive type [" + archiveType + "] is not supported.");
```

pickypg (commented):
One real comment and one question.

src/main/java/com/elastic/support/monitoring/MonitoringExportInputs.java
brunofarache: Done.

This PR also fixes a nasty bug. When `port` is invalid, many port validation errors are printed out. This was happening because `emptyList` is mutable and shared. I have another PR in the queue to avoid those issues in the future.

src/main/java/com/elastic/support/monitoring/MonitoringExportService.java
brunofarache: Done.

It's weird, IntelliJ keeps adding them back automatically.

pickypg: IntelliJ has pretty low defaults for doing that. Most teams I work on suggest setting the limit to 9999.

pickypg (approved):
LGTM

PR #506, opened 2021-07-09
Apply missing pretty formats

stefnestor:
Adds in missing columns, pretty json, & sorting onto existing request uri's.

### DIFF
--- a/src/main/resources/elastic-rest.yml
+++ b/src/main/resources/elastic-rest.yml
@@ -32,7 +32,7 @@ cat_count:
   extension: ".txt"
   subdir: "cat"
   versions:
-    ">= 0.9.0": "/_cat/count"
+    ">= 0.9.0": "/_cat/count?v"
 
 cat_fielddata:
   extension: ".txt"
@@ -59,7 +59,7 @@ cat_master:
   extension: ".txt"
   subdir: "cat"
   versions:
-    ">= 0.9.0": "/_cat/master"
+    ">= 0.9.0": "/_cat/master?v"
 
 cat_nodeattrs:
   extension: ".txt"
@@ -71,7 +71,7 @@ cat_nodes:
   extension: ".txt"
   subdir: "cat"
   versions:
-    ">= 0.9.0": "/_cat/nodes?v&h=n,nodeId,m,i,r,d,hp,rp,cpu,load_1m,load_5m,load_15m,"
+    ">= 0.9.0": "/_cat/nodes?v&s=master,name&h=n,nodeId,m,i,r,d,hp,rp,cpu,load_1m,load_5m,load_15m,"
 
 cat_pending_tasks:
   extension: ".txt"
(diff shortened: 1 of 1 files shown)


src/main/resources/elastic-rest.yml
```diff
@@ -139,7 +139,7 @@ alias:
 
 allocation:
   versions:
-    ">= 0.9.0": "/_cat/allocation?v&format=json"
+    ">= 0.9.0": "/_cat/allocation?v&pretty&format=json"
```
pickypg, line 142: Have you validated that these work with older versions?

inqueue:
My vote is for less `?pretty` since it adds extra overhead to diagnostic collection and bundle size. +1 to removing it whenever we can.

src/main/resources/elastic-rest.yml
stefnestor: I did not 😞 - AFAICT `pretty` is used / ignored but doesn't cause calls to fail. Is that incorrect?

stefnestor:
> My vote is for less ?pretty since it adds extra overhead to diagnostic collection and bundle size. +1 to removing it whenever we can.

@inqueue, do we have an alternative way of formatting all JSON on save, because agreed that would be preferred. This is just bringing formatting to expected par through current method.

inqueue:
> > My vote is for less ?pretty since it adds extra overhead to diagnostic collection and bundle size. +1 to removing it whenever we can.
> 
> @inqueue, do we have an alternative way of formatting all JSON on save, because agreed that would be preferred. This is just bringing formatting to expected par through current method.

I don't other than the consumer using `jq` or other tools to prettify. The problem is we don't know when pretty json is going to break a tool, https://github.com/elastic/support-diagnostics/pull/468 for example.

API call output size for those in this PR are typically nowhere near what we might see in a cluster state. My comment was aimed at keeping consistency in data collection across all API calls, but it can come at a cost of interrupting workflow, requiring additional tools, etc. I'm fine with adding pretty to these calls...or not :)

jpcarey:
my personal opinion is to not have the `?pretty` flag. I think we should do our best to reduce our footprint at collection time (size, memory, etc). Having a simple tool / script to "extract" the data on our side could take care of prettifying the json, and we could reduce the number of API calls (eg. the cluster state contains the templates and mappings, why fetch those separately).

stefnestor:
>  I think we should do our best to reduce our footprint at collection time (size, memory, etc). Having a simple tool / script to "extract" the data on our side could take care of prettifying the json

@jpcarey IMO "on our side" should still be within the diagnostic code and not a separate Elastic tool, so that the original diagnostic pull is as helpful as possible to the widest audience. I agree on having a lighter footprint & believe switching out `?pretty` for a simple JSON formatting code would be extremely light within the diagnostic itself & lighter than it currently is.

This would still allow the diagnostic to be usable by Elastic Support as well as Elastic users & not restrict widespread audience usage or require a tool outside the diagnostic itself to digest it. I'm not familiar w/how to add separate sub-code paths in terms of this repo, so I'll close this PR & file as an issue. Thanks!

PR #514, opened 2021-08-07
add new API and spaces management for Kibana diagnostic v2

CamiloSierraH:
Previous PR was summited 3 months ago, and there were some changes on the code that created conflicts.
I have fixed the conflicts and added the code that will collect the alerts and health for all the spaces.

cc @pickypg @danleerunk

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
     <groupId>co.elastic.support</groupId>
     <artifactId>diagnostics</artifactId>
-    <version>8.2.2</version>
+    <version>8.2.1-SNAPSHOT</version>
     <packaging>jar</packaging>
     <name>Support Diagnostics Utilities</name>
     <description>Elastic Support Diagnostics Utilities</description>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
     <groupId>co.elastic.support</groupId>
     <artifactId>diagnostics</artifactId>
-    <version>8.2.1-SNAPSHOT</version>
+    <version>8.2.2</version>
     <packaging>jar</packaging>
     <name>Support Diagnostics Utilities</name>
     <description>Elastic Support Diagnostics Utilities</description>
--- a/src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
+++ b/src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
@@ -28,6 +28,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Iterator;
 
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.io.FileWriter;
 import java.io.IOException;
 
@@ -323,8 +325,8 @@ public class RunKibanaQueries extends BaseQuery {
 
                     // If any headers were removed, we need to rewrite the file to remove them
                     if (headerRemoved == true) {
-                        String fileName = context.tempDir + SystemProperties.fileSeparator + "kibana_actions_"+ spaceId +".json";
-                        try (FileWriter fileWriter = new FileWriter(fileName)) {
+                        Path path = Paths.get(context.tempDir, "kibana_actions_" + spaceId + ".json");
+                        try (FileWriter fileWriter = new FileWriter(path.toFile())) {
                             fileWriter.write(actions.toPrettyString());
                             fileWriter.flush();
                         } catch (IOException e) {
--- a/src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
+++ b/src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
@@ -98,7 +98,7 @@ public class RunKibanaQueries extends BaseQuery {
     * @param  perPage  Number of docusment we reques to the API
     * @param  action Kibana API name we are running
     */
-    private void runActionsBySpace(RestClient client, List<RestEntry> queries, int perPage, RestEntry action) {
+    public void runActionsBySpace(RestClient client, List<RestEntry> queries, int perPage, RestEntry action) throws DiagnosticException {
         List<String> spaces = getAllSpaces(client);
         String url = action.getUrl();
         String baseUrl = url.substring(4, url.length());
@@ -113,7 +113,11 @@ public class RunKibanaQueries extends BaseQuery {
             RestEntry copyAction = (RestEntry) action.clone();
 
             if (copyAction.getUrl().contains("/_find")) {
-                getAllPages(client, queries, perPage, copyAction, spaceId);
+                try {
+                    getAllPages(client, queries, perPage, copyAction, spaceId);
+                } catch (DiagnosticException e) {
+                    logger.error(String.format("Space: %s has no pagination", spaceId));
+                }
             } else {
                 String actionName = baseName + "_" + spaceId;
                 copyAction.setName(actionName);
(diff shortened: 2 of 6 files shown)


src/main/java/co/elastic/support/rest/RestEntry.java
```diff
@@ -36,6 +36,14 @@ public String getUrl() {
         return url;
     }
 
+    public void setUrl(String url) {
```
pickypg, line 39: I cannot remember if we discussed this last time or not, but these methods change this class from being immutable to mutable, which is rarely a good idea. As a one-off change, it's honestly probably safe, but if we ever changed to multithread the REST calls, then it would become risky.

The better approach here is a mix of what you're doing: replace the usage of `clone` with another function whose purpose is less generic, but functionally equivalent.

```java
/**
 * Change the name and URL of the current {@code RestEntry}.
 *
 * @param name The new name to use
 * @param url The new URL to use
 * @return A copy of the current {@code RestEntry} with the {@code name} and {@code url}.
 */
public RestEntry withNameAndUrl(String name, String url) {
    return new RestEntry(name, getSubdir(), getExtension(), isRetry(), url, showErrors);
}
```

This one-line function keeps the code immutable, makes the caller code easier, and avoids using `clone`, which requires a cast.

pom.xml
```diff
@@ -4,7 +4,7 @@
 
     <groupId>co.elastic.support</groupId>
     <artifactId>diagnostics</artifactId>
-    <version>8.2.1-SNAPSHOT</version>
+    <version>8.2.2</version>
```
pickypg, line 7: Since 8.2.1 is unreleased, the version can stay the same.

```suggestion
    <version>8.2.1-SNAPSHOT</version>
```

pickypg (changes requested):
Minor issues, mostly nitpicks.

src/main/resources/kibana-rest.yml
```diff
@@ -22,6 +22,10 @@ kibana_alerts:
   versions:
     ">= 7.9.0": "/api/alerts/_find"
 
+kibana_alerts_health:
+  versions:
+    ">= 7.11.0": "/api/alerts/_health"
```
stefnestor, line 27: Is this deprecated now (from 7.13) per [this doc](https://www.elastic.co/guide/en/kibana/current/alerts-api.html#alerts-api-health)?

src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -96,27 +137,71 @@ public int runBasicQueries(RestClient client, DiagnosticContext context) throws
...
+                            if (!(name.equals(null) || name.equals(""))) {
+                                this.allSpaces.add(name.replace("\"",""));
+                            }
+                        }
+                    }
                 }
             }
+            //if we can not find any space, set the default space id
+            if (this.allSpaces.size() <= 0) {
+                this.allSpaces.add("default");
```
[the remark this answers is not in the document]
stefnestor: @CamiloSierraH, I believe this PR is blocked waiting update on @pickypg's feedback/questions.

@pickypg, I agree I interpret the same from [this doc](https://www.elastic.co/guide/en/kibana/current/spaces-api-get-all.html#spaces-api-get-all-example-1) which shows `default` explicitly returns. Supposedly `default` can't be `id` edited nor deleted which is why @CamiloSierraH has the catch, but I'd think if that magically happened it'd be noteworthy in itself.

src/main/resources/kibana-rest.yml
stefnestor: Bump @CamiloSierraH 🙏🏼

pickypg:
Closing this one out. We haven't had a need for it since the time it was created (worked well) and the changes to paging mutability seem undesirable in the long run.

stefnestor:
> We haven't had a need for it since the time it was created (worked well)

@pickypg I'm on board we may need to do this another way, but I'm eagerly waiting when we finally pull Rules from non-default Space which this PR enables. Also, if we need to send this back to Kibana Dev, happy to figure out how to request from them (@pmuellr in case he knows since he knows everything). 

But I don't think this is ok to remain as is, not since Rules became Space specific/aware, because we don't pull, Rules from non-default Spaces when they exist. E.g. SIEM use-cases IME always push users to work out of the non-default Space & w/o this PR we're always flying blind on those SIEM Spaces.

<details><summary>Current example</summary>

1. Set up a Kibana cluster with two spaces [`default`, `space2`]
    image
1. Create a single Rule in `space2` titled `test1`
    image
1. Pull Kibana diagnostic, noting no `kibana_alerts_1.json` file.
    image
1. Create a second Rule in `default` titled `test2`
    image
1. Pull Kibana diagnostic, noting only a single rule `test2` under `kibana_alerts_1.json`
    image
1. This indicates the Kibana diagnostic only covers the `default` space at least for Rules & Rules in every other Space are missed and therefore the diagnostic is incomplete.

</details>

cc: @danleerunk

pmuellr:
Ya, it's unfortunate we only get the default space rules - a brief peek at the code indicates we are querying Kibana for these, but it would be a lot easier to query ES for them instead - we can get all the rules | connectors for all spaces in one call.

Please hit us up in #kibana-alerting if you have any questions on this stuff.

BTW, I recently created an issue https://github.com/elastic/kibana/issues/124382 to do some research on making alerting more observable **specifcally** for cloud, and one of the thoughts was to enhance the support-diagnostics, to look for "troublesome" things: error counts, longest running rules, etc.  So I suspect we'll be back around here, at some point :-).  The kind of queries we'd do for that would be ES queries against the `.kibana-event-log*` indices.

pickypg:
In order to get those into the support diagnostic, the best approach would be to provide those behind a set of diagnostic-friendly Kibana APIs that can be invoked. That way as the backing indices change (or access to them, as system indices changes), there's no risk to the diagnostic needing to understand different versions and there's no risk of it slurping up user info versus diagnostic info.

stefnestor:
From https://github.com/elastic/support-diagnostics/pull/541, AFAIK we do not add `.kibana`, `.kibana-event-log*` queries within either Elasticsearch/Kibana diagnostics. AFAICT we need to separately tackle the initial above rules from all spaces & then after tackle a more automated way to pull expensive queries (which I agree nicely summarizes under https://github.com/elastic/kibana/issues/124382).

I think the question/ask is, 1) is there a way to pull all rules from all spaces in one query (and not iterate over spaces) or 2) how can we best request Kibana Dev assistance to expand the diagnostic to iterate over spaces so we can get the information we need from the diagnostic? Or, maybe there's a better way to go about this :)

pmuellr:
> 1) is there a way to pull all rules from all spaces in one query (and not iterate over spaces)

Not with the Kibana APIs, only ES APIs.  Using ES would be preferred anyway, since the data returned from the Kibana Alerting APIs is slightly different (removed properties) than what we actually store in ES.  I'd think we'd want raw ES access anyway so we could do some aggs on # of visualizations/etc, which is probably not possible with the Kibana APIs.  And certainly interesting queries over the event log are only possible via ES, since the Kibana APIs only expose a limited amount of information from the event log.

I assume that impacts the tooling, since Kibana doesn't expose the ES URL via it's APIs, so you'd have to pass both the ES and Kibana URLs to the kibana diagnostics tool.  

> 2) how can we best request Kibana Dev assistance to expand the diagnostic to iterate over spaces so we can get the information we need from the diagnostic?

You can get the list of spaces via the spaces API: https://www.elastic.co/guide/en/kibana/current/spaces-api-get-all.html

stefnestor:
I can't personally fix this & need to compensate for it weekly, so have moved the conversation to https://github.com/elastic/support-diagnostics/issues/578 so we don't loose it. Cheers!

PR #519, opened 2021-08-25
Add support for additional headers to DiagnosticService

antbell:
Add support to specify additional request headers to be passed when making API requests to the Elasticsearch server.
This is needed in some environments where custom headers may be required for authorization purposes, logging etc.


pickypg (approved):
LGTM

PR #522, opened 2021-09-06
Update license header to Elastic License v2

brunofarache:
The source header was wrong, it was using the unversioned header. It should explicitly say it's ELv2.

Also updated `NOTICE.txt` manually to make license names consistent and added the full text of some.


brunofarache:
cc @pickypg @antbell

pickypg (approved):
LGTM

PR #523, opened 2021-09-08
double ampersand in cat_indices call

eedugon:
i have included the s=index which i guess it was missing, and also fixed the double ampersand that didn't look good


nellicus (approved):
LGTM

PR #533, opened 2021-09-16
Add textIOManager to add parseInputs

pickypg:
This adds `@Override` to all `parseInputs` methods to guarantee that they're overriding the root implementation rather than serving as overloads that were going uninvoked.

The root failure of _not_ doing that was that inputs that required user inputs were not triggered (e.g., `-p`).

This also fixes a typo in the ScrubApp's version.


antbell (approved):
LGTM

jpcarey (approved):

PR #541, opened 2021-10-02
+IndexPattern API pull

stefnestor:
👋🏼 I would like to recommend automating pulling Kibana's Index Pattern list for your consideration. While the feature fits better under Kibana, I've figured out to pull it via Elasticsearch API & therefore have put it under that code area. Thanks in advance!

Key investigation benefits

- are user requests for index / index pattern
- what timeFieldName is used
- answering column differences between index mappings & why user thinks "works sometimes & not others" aka. based on time range back

image

### DIFF
--- a/src/main/resources/elastic-rest.yml
+++ b/src/main/resources/elastic-rest.yml
@@ -203,6 +203,10 @@ index_templates:
   versions:
     ">= 7.8.0": "/_index_template"
 
+index_patterns:
+  versions:
+    ">= 0.9.0": "/.kibana/_search?size=100&q=type:index-pattern"
+
 indices:
   versions:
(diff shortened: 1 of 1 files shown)


pickypg (commented):
I do not think that we can / should do this. We currently tell our users that the ES diagnostic does not read their data, and this would invalidate that statement.

Separately, the Kibana diagnostic should help with some of this.

stefnestor:
Noting for self, Kibana does have an [endpoint for this](https://www.elastic.co/guide/en/kibana/current/data-views-api-get.html), but it's space deliminated requiring https://github.com/elastic/support-diagnostics/issues/578

PR #551, opened 2021-12-06
Add leading / to elastic rest endpoints which were missing it.

antbell:
## Description

The two ES endpoints related to transforms are not retrieved properly when the tool is used in library-mode. It turns out that they are the only ones missing a leading / in the configuration file `elastic-rest.yml`.
This PR adds the leading / and adds a related check to `TestRestConfigFileValidity.java`.


pickypg (approved):
LGTM

PR #558, opened 2021-12-17
feature: retrieve disk usage via _disk_usage API

ilvar:
Add the new [_disk_usage](https://www.elastic.co/guide/en/elasticsearch/reference/7.15/indices-disk-usage.html) endpoint for Elasticsearch

### DIFF
--- a/src/main/resources/elastic-rest.yml
+++ b/src/main/resources/elastic-rest.yml
@@ -189,6 +189,10 @@ dangling_indices:
   versions:
     ">= 7.9.0": "/_dangling?pretty&human"
 
+disk_usage:
+  versions:
+    ">= 7.15.0": "/*/_disk_usage?run_expensive_tasks=true&flush=false&ignore_unavailable=true"
+
 fielddata:
   versions:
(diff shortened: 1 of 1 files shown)


pickypg:
I created an issue with the ES team to see if they might consider a `GET` request variant that drops the `flush` parameter entirely.

https://github.com/elastic/elasticsearch/issues/82594

pickypg (changes requested):
Blocked by https://github.com/elastic/elasticsearch/issues/82594.

ilvar:
@pickypg I've signed CLA after creating the PR though looks like it's not propagated. Re-signed today; double checked than it matches full name and email in my git config

pickypg:
Thanks @ilvar. It looks like the CLA checker is happy now. Now we just have to wait for the ES team to decide if they want to add a `GET` version of this API and, if so, one of us will need to tweak the version(s).

pickypg:
It looks like the ES team does not want to add this API as a `GET`. Closing this one out. Thanks for the try though!

PR #563, opened 2022-01-09
close #562 - fix message when running bash file from sources

### LINKED ISSUE #562: running diagnostics.sh from source code does not send clear message
I think code https://github.com/elastic/support-diagnostics/pull/374 never worked for bash version

1- Download latest release source file (I know this is added by GH when creating a release and we can't disable this)
2- unzip/untar and run `scripts/diagnostics.sh` : 
```
$ scripts/diagnostics.sh
Using /usr/local/opt/openjdk@8/bin/java as Java Runtime
Error: Could not find or load main class co.elastic.support.diagnostics.DiagnosticApp
```

While I'm there, there is another issue in the code [here](https://github.com/elastic/support-diagnostics/blob/v8.2.6/scripts/diagnostics.sh#L7) : 
```
if [  -d "libDir" ]; then
```

This is always false (there is never a directory with explicit value `libDir`, it should read
```
if [ ! -d "$libDir" ]; then
```

After this I get the expected behaviour: 
```
$ ./diagnostics.sh
Runtimes library does not exist - make sure you are running the
archive with 'dist-' in the name, not the one labeled: 'source'.
```

Also unrelated to be clear in bash the way to use a variable is `"$variable"` - the double-quote are also mandatory otherwise if the value contain a space it's not likely to work for some commands (it will replace value with `/some/folder with space` where the value should be `/some/folder\ with\ space` or `"some/folder with space"`... This does not seem to cause any issue in this case for -d check though it's not very clean code
I'll send a PR for this

jguay:
This code fixes incorrect check for lib directory for `diagnostics.sh`

I also added double quotes for variables that contain a path to prevent globing and word splitting.

### DIFF
--- a/scripts/diagnostics.sh
+++ b/scripts/diagnostics.sh
@@ -5,9 +5,13 @@ scriptDir="${scriptDir/\/diagnostics.sh/$''}"
 libDir="$scriptDir"'/lib'
 
 if [ ! -d "$libDir" ]; then
-    echo "Runtimes library does not exist - make sure you are running the "
-    echo "archive with 'dist-' in the name, not the one labeled: 'source'."
-    exit
+    echo "Diagnostic executable not found:"
+    echo ""
+    echo "Please make sure that you are running with the archive ending with"
+    echo "'-dist.zip' in the name and not the one labeled 'Source code'."
+    echo ""
+    echo "Download at https://github.com/elastic/support-diagnostics/releases/latest"
+    exit 400
 fi
 
 if [ -x "${JAVA_HOME}/bin/java" ]; then
--- a/scripts/diagnostics.sh
+++ b/scripts/diagnostics.sh
@@ -1,17 +1,17 @@
 #!/usr/bin/env bash
 
-scriptDir=$0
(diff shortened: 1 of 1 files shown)


scripts/diagnostics.sh
```diff
@@ -1,17 +1,17 @@
...
-libDir=$scriptDir'/lib'
+scriptDir="$0"
+scriptDir="${scriptDir/\/diagnostics.sh/$''}"
+libDir="$scriptDir"'/lib'
 
-if [  -d "libDir" ]; then
+if [ ! -d "$libDir" ]; then
     echo "Runtimes library does not exist - make sure you are running the "
     echo "archive with 'dist-' in the name, not the one labeled: 'source'."
     exit
```
pickypg, line 10:
```suggestion
    echo "Diagnostic executable not found:"
    echo ""
    echo "Please make sure that you are running with the archive ending with"
    echo "'-dist.zip' in the name and not the one labeled 'Source code'."
    echo ""
    echo "Download at https://github.com/elastic/support-diagnostics/releases/latest"
    exit 400
```

pickypg (approved):
LGTM with an update to the message.

PR #567, opened 2022-01-26
Add list SLM repositories

stefnestor:
@pickypg @dakrone may we add the `GET _snapshot` request to the ES diag? Currently we return `GET _cat/repositories` but it doesn't include sub information used for troubleshooting SLM repositories. Not interpreting returned data as user private but kindly decline if it could return sensitive data. 🙏🏼

### DIFF
--- a/src/main/resources/elastic-rest.yml
+++ b/src/main/resources/elastic-rest.yml
@@ -269,6 +269,10 @@ remote_cluster_info:
   versions:
     ">= 6.0.0": "/_remote/info"
 
+repositories:
+  versions:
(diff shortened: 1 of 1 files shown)


dakrone:
It sounds reasonable to me, but I'm not sure what the line for sensitive vs non-sensitive is.

pickypg:
We try to avoid fetching user data. Snapshots are right on that line, similar to index names and mappings. However, I think it strikes the right balance between fetching necessary information to debug things and not overstepping.

pickypg (approved):
LGTM

PR #568, opened 2022-01-27
changed  version check error message to warning.

TomonoriSoejima:
A user ran into this message `was bypassed due by version check` and asked me if this was of any concern.
I think the log type can be relaxed to `warn` perhaps?


nellicus (approved):
LGTM

PR #572, opened 2022-02-10
Read Kibana version from /api/stats instead of /api/settings

### LINKED ISSUE #571: Fix version detection for Kibana 8.0
Kibana 8.0 doesn't expose the `api/settings` route, which is used to detect the Kibana version before calling the other rest endpoints.
The Kibana version can be read from the payload of the `/api/stats` route.

antbell:
## Description

Fixes #571


pickypg (approved):
LGTM. Thanks for tackling this @antbell!

PR #576, opened 2022-02-28
Retrieve top-level data_stream property from indices

VimCommando:
We currently collect `alias`, `settings` and `mappings` but the top-level `data_stream` index property isn't in the diagnostic anywhere. Having this will greatly help automate and visualize data streams in our analyzer.

PR #577, opened 2022-03-01
Collect all data streams (including system/hidden ones)

lucabelluccini:
Adjust the collection of `_data_stream` to get also hidden/system indices.

This is particularly useful for 8.0 where some system data is in hidden/system data streams instead of normal indices

### DIFF
--- a/src/main/resources/elastic-rest.yml
+++ b/src/main/resources/elastic-rest.yml
@@ -515,3 +515,4 @@ data_stream:
   subdir: "commercial"
   versions:
     ">= 7.9.0": "/_data_stream?pretty"
+    ">= 7.11.0": "/_data_stream?pretty&expand_wildcards=all"


pickypg:
@lucabelluccini @VimCommando How does this compare to #576? They seem to be trying to get the same information in different ways (and this one has the benefit of a single file, but the other almost certainly collects more info).

lucabelluccini:
Hello @pickypg
I think the differences are the `expand_wildcards` on DS API has been added only in 7.11.
On 7.9 we get:
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "request [/_data_stream] contains unrecognized parameter: [expand_wildcards]"
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "request [/_data_stream] contains unrecognized parameter: [expand_wildcards]"
  },
  "status" : 400
}

```

pickypg:
In that case, I think we want this one. Thanks @lucabelluccini

Issue #578, opened 2022-03-02
Enable Kibana Diagnostic to be Space Aware for Rules


stefnestor:
Consolidating parallel conversations in this repository (https://github.com/elastic/support-diagnostics/issues/479, https://github.com/elastic/support-diagnostics/pull/514, https://github.com/elastic/support-diagnostics/issues/512, https://github.com/elastic/kibana/issues/124382). The Kibana diagnostic is not currently space aware though Kibana (SIEM) Rules have been space delineated since at least 7.10. 

This implicates that the Kibana diagnostic only captures (SIEM) Rule info from `space:default` & misses any/all other spaces. (You can determine this from an export because the [Task Manager Health API](https://www.elastic.co/guide/en/kibana/current/task-manager-health-monitoring.html) will have scheduled Rule counts surpassing your exported Rule counts.) 

SIEM appears more highly impacted because industry Security practices frequently suggest avoiding allocating users to a product's default space rather than purposely creating a new non-default space for them.

Example here: https://github.com/elastic/support-diagnostics/pull/514#issuecomment-1027404877.

cc: @pmuellr @pickypg @CamiloSierraH @111andre111 @mikecote

pmuellr:
Heh, was just looking at this yesterday.

Ideally, we could get the alerting connectors / rules directly from ES, and not Kibana.  But I think the problem with that approach is the diagnostic tool either points to ES when getting ES diags, or Kibana when getting Kibana diags, so when running the Kibana diags, the ES url is not available.

I say "ideally", but honestly not sure - there may be some reason it would be preferable to get the data from Kibana.  I think the ES versions will likely have "more stuff" than the Kibana versions, which seems like a good thing.  @mikecote ?

If we want to continue down the path of just using the Kibana URL, we'll need to get all the spaces, then for each space, do what's currently being done with the `default` space.

At some point, connectors / rules will be shareable across spaces, so I guess if we continue down the path of using the Kibana APIs, we'd end up having dups of rules/connectors that are shared.  Not a huge problem, but one we won't have if we query the rules right from ES.

pickypg:
> I say "ideally", but honestly not sure - there may be some reason it would be preferable to get the data from Kibana. I think the ES versions will likely have "more stuff" than the Kibana versions, which seems like a good thing. @mikecote ?

[...]

Issue #479, opened 2021-04-14
Kibana diag improvements for prepackaged, stats, task health and new alerts health APIs


kostasb:
1) The `/api/detection_engine/prepackaged` Kibana API returns HTTP 404 with 7.11 and 7.12. We may want to remove this call from the diag tool

```
kibana_detection_engine_signals:
  versions:
    "> 7.10.0": "/api/detection_engine/prepackaged"
```

2) For the stats endpoint it is preferrably to retrieve the extended version which is much more verbose.
```
kibana_stats:
  versions:
    ">= 6.5.0": "/api/stats"
```

instead use: `/api/stats?extended=true`

3) The task manager health api is named "kibana_task_manager_help" in the diag tool, which generates output file "kibana_task_manager_help.json", it should better be "health" instead of "help" to align with the API call

```
kibana_task_manager_help:
  versions:
    ">= 7.11.0": "/api/task_manager/_health"
```

4) As of 7.11 there is a new Alert Framework Health API https://www.elastic.co/guide/en/kibana/current/alerts-api-health.html

Can we add this new API, i.e.:
```
kibana_alerts_health:
  versions:
    ">= 7.11.0": "/api/alerts/_health"
```

@CamiloSierraH what do you think?

CamiloSierraH:
Hey @kostasb and @danleerunk I will work on this now, I will submit a new PR asap!

CamiloSierraH:
I finished the first 3 changes.
The last one is interesting, by default only get the data for the default space, if the customer have multiple spaces, we will not collect all the information, I will do a code and review the other queries we have to be sure that we collect not only the data in default space, but in all the spaces.

```
GET <kibana host>:<port>/api/alerting/_health

GET <kibana host>:<port>/s/<space_id>/api/alerting/_health
```

CamiloSierraH:
I need to test it first but the code is here: https://github.com/CamiloSierraH/support-diagnostics/commit/8473f39e37aec9c366d3c9434942d3dcb43cce65

I updated all the kibana APIs (alerts, actions, detection_engine), so now we are collecting all the spaces data.
For the new query `kibana_alerts_health` we do the same, and we collect all the spaces data!

Hey @kostasb my friend if you want to test it you can see here the new jar https://drive.google.com/file/d/1LGq990SU1loTZj_bMuInyXEcfqV78eTU/view?usp=sharing

I need to run some test with kibana in remote and locat before push the PR, but I will push it tomorrow morning!

[...]

PR #580, opened 2022-03-03
Append Kibana Security APIs

stefnestor:
@pmuellr @jportner to better introspect Kibana Security/Space issues, adds APIs into Kibana diagnostic: 
- [List Kibana Roles](https://www.elastic.co/guide/en/kibana/current/role-management-api-get.html)
- [List Kibana Spaces](https://www.elastic.co/guide/en/kibana/current/spaces-api-get-all.html)
    - Enables some introspection to mitigate [this](https://github.com/elastic/support-diagnostics/issues/578)'s impact.
- Get Current Kibana User
    - URL marks internal, but this is documented [here](https://github.com/elastic/support-diagnostics/issues/479#issuecomment-825148583) & is frequently used to diagnose Kibana permission issues.

Kindly review
1. If endpoints return user private/sensitive data.
2. Initial available versions.
    - I pulled via Github search / oldest doc, but there's been a lot of migrations to pinpoint oldest well.

cc: @pickypg

### DIFF
--- a/src/main/resources/kibana-rest.yml
+++ b/src/main/resources/kibana-rest.yml
@@ -40,7 +40,7 @@ kibana_roles:
 
 kibana_spaces:
   versions:
-    ">= 6.4.0": "/api/spaces/space?include_authorized_purposes=true"
+    ">= 6.5.0": "/api/spaces/space"
 
 kibana_settings:
   versions:
@@ -57,3 +57,4 @@ kibana_task_manager_help:
 kibana_user:
   versions:
     ">= 7.6.0": "/internal/security/me"
+    ">= 5.0.0 < 7.6.0": "/api/security/v1/me"
--- a/src/main/resources/kibana-rest.yml
+++ b/src/main/resources/kibana-rest.yml
@@ -34,6 +34,14 @@ kibana_detection_engine_signals:
   versions:
     "> 7.10.0": "/api/detection_engine/prepackaged"
 
+kibana_roles:
+  versions:
+    ">= 6.4.0": "/api/security/role"
+
+kibana_spaces:
+  versions:
+    ">= 6.4.0": "/api/spaces/space?include_authorized_purposes=true"
+
 kibana_settings:
   versions:
     ">= 6.5.0 < 8.0.0": "/api/settings"
@@ -45,3 +53,7 @@ kibana_stats:
 kibana_task_manager_help:
   versions:
     ">= 7.11.0": "/api/task_manager/_health"
+
+kibana_user:
+  versions:
+    ">= 7.6.0": "/internal/security/me"


pmuellr:
I'm just seeing a single commit 7cd356748db12c84ddce6c5b9f219186b47b2665 with a change to a yaml file.  Is there supposed to be some code here as well?

jportner:
* **List Kibana roles**: This should be OK, it doesn't return private data. This was introduced in #19723, initial available version is 6.4.0.
* **List Kibana spaces**: This should be OK but it's not clear to me how having this information will assist in troubleshooting the linked issue. Can you clarify what the expectation here is?
* **Get Current Kibana User**: This can contain some "private data" in terms of personal information, example:
   ```json
   {
     "username": "elastic",
     "roles": [
       "superuser"
     ],
     "full_name": null,
     "email": null,
     "metadata": {
       "_reserved": true
     },
     "enabled": true,
     "authentication_realm": {
       "name": "reserved",
       "type": "reserved"
     },
     "lookup_realm": {
       "name": "reserved",
       "type": "reserved"
     },
     "authentication_type": "realm",
     "authentication_provider": {
       "type": "basic",
       "name": "basic"
     }
   }
   ```
   If the user is authenticated via SSO and their ES is configured to map attributes correctly, the `full_name` and/or `email` fields can be populated. That said, the rest of this data is really valuable for troubleshooting, and I think we can justify collecting this small amount of personal information for a support case where we likely already have this info anyway.
   * `/internal/security/me`: this was introduced in #50695, initial available version is 7.6.0.
   * `/api/security/v1/me`: this API gives the same info, it is deprecated and is being replaced by the new API above; I believe it was introduced in https://github.com/elastic/x-pack/commit/003ef56664a74bbb31d23ada3f24895ff916502e, initial available version appears to be 5.0.0.

pmuellr:
> List Kibana spaces: This should be OK but it's not clear to me how having this information will assist in troubleshooting the linked issue. Can you clarify what the expectation here is?

Currently, the diagnostics get the alerting rules and connectors from the default space only.  We'd like to have them for all spaces, so presumably the space names will end up showing up here.   The diagnostics code will eventually be changed to get the list of spaces, and for each one, get all the rules and connectors from each space.

jportner:
> Currently, the diagnostics get the alerting rules and connectors from the default space only. We'd like to have them for all spaces, so presumably the space names will end up showing up here. The diagnostics code will eventually be changed to get the list of spaces, and for each one, get all the rules and connectors from each space.

Gotcha, that makes sense.

@stefnestor the List Kibana Spaces API was introduced in https://github.com/elastic/kibana/pull/21408, initial available version is 6.5.0.
However, the `include_authorized_purposes` parameter wasn't added until 7.11.0. I would omit that from your test script entirely, it won't give you useful information.

src/main/resources/kibana-rest.yml
```diff
@@ -34,6 +34,14 @@ kibana_detection_engine_signals:
   versions:
     "> 7.10.0": "/api/detection_engine/prepackaged"
 
+kibana_roles:
+  versions:
+    ">= 6.4.0": "/api/security/role"
+
+kibana_spaces:
+  versions:
+    ">= 6.4.0": "/api/spaces/space?include_authorized_purposes=true"
```
jportner, line 43:
```suggestion
    ">= 6.5.0": "/api/spaces/space"
```

src/main/resources/kibana-rest.yml
```diff
@@ -45,3 +53,7 @@ kibana_stats:
 kibana_task_manager_help:
   versions:
     ">= 7.11.0": "/api/task_manager/_health"
+
+kibana_user:
+  versions:
+    ">= 7.6.0": "/internal/security/me"
```
jportner, line 59:
```suggestion
    ">= 7.6.0": "/internal/security/me"
    ">= 5.0.0 < 7.6.0": "/api/security/v1/me"
```

stefnestor:
Thanks, both 🎉 ! Late circling back, but changes accepted & ready for Dev approval for SupportOps to then merge. 🙏🏼

pickypg (approved):
LGTM

PR #589, opened 2022-04-30
Add Runner to Manifest

pickypg:
This adds the concept of a `runner` to indicate who or what ran the diagnostic, and adds it to the `manifest.json` file.

It also fixes a few things, such as forcing the UTC date functions to actually force UTC, as well as printing the ISO-8601 formatted timezone instead of the RFC 822 ones.

Finally, it fixes a malformed ES action (because it should match one and only one API per action) and removes some Kibana tests that did not reflect the current state of output.

### DIFF
--- a/src/test/java/co/elastic/support/diagnostics/commands/TestRunKibanaQueries.java
+++ b/src/test/java/co/elastic/support/diagnostics/commands/TestRunKibanaQueries.java
@@ -40,7 +40,7 @@ import static org.mockserver.model.HttpResponse.response;
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class TestRunKibanaQueries {
 
-	private static final Logger logger = LogManager.getLogger(RestClient.class);
+    private static final Logger logger = LogManager.getLogger(RestClient.class);
     private ClientAndServer mockServer;
     private RestClient httpRestClient, httpsRestClient;
     private String temp = SystemProperties.userDir + SystemProperties.fileSeparator + "temp";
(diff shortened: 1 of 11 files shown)


src/test/java/co/elastic/support/diagnostics/commands/TestRunKibanaQueries.java
```diff
@@ -90,300 +90,96 @@ public void tearDown() {
...
+        RestEntryConfig builder = new RestEntryConfig(version);
 
         Map restCalls = JsonYamlUtils.readYamlFromClasspath(Constants.KIBANA_REST, true);
         Map<String, RestEntry> entries = builder.buildEntryMap(restCalls);
-    	context.elasticRestCalls = entries;
-
-    	context.tempDir = tempDir.getPath();
-    	context.perPage = 2;
-    	
-        mockServer
```
pickypg, line 106: This boilerplate is all replaced with `configMockServerRoute` above.

src/test/java/co/elastic/support/diagnostics/commands/TestRunKibanaQueries.java
```diff
@@ -40,7 +40,7 @@
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
pickypg, line 40: This file is test cleanup related to recently added Kibana actions.

jpcarey (approved):
LGTM.  Fetched an Elasticsearch diagnostic (`--type api`) and verified the `runner` value in the `manifest.json`.

PR #594, opened 2022-05-23
add security API calls

### LINKED ISSUE #512: endpoint security api calls for elasticsearch and kibana diagnostics:
[1,2,2,1]

Due to the upcoming version 7.14.0, support diagnostics should now obtain following api entries:
Kibana:
```
Endpoint Metadata            : /api/endpoint/metadata/{agent.id}
Endpoint Policy              : /api/fleet/package_policies/{Endpoint.policy.applied.id}
Endpoint Policy response     : /api/endpoint/policy_response?agentId={agent.id}
Security packages data       : /api/fleet/epm/packages?experimental=true&category=security
Trusted apps                 : /api/endpoint/trusted_apps
Associated Fleet Agent status: /api/fleet/agents/{elastic.agent.id}
Agent status 2               : /api/endpoint/action_status?agent_ids={elastic.agent.id}
Transforms                   : /api/transform/transforms
Transform status/logs for the endpoint.metadata_current transform: /api/transform/transforms/{transform.name}/messages  transform.name e.g. endpoint.metadata_current-default-0.20.2
Transform stats logs         : /api/transform/transforms/_stats
Associated Fleet Agent Logs  : /api/fleet/agent_policies/{agent.id} und .item.package_policies   and then  /api/fleet/epm/packages/{package.name}  e.g. system-1.0.0   or --> They can also be access via Logs app with the following filter: elastic_agent.id:{elastic.agent.id} and (data_stream.dataset:elastic_agent)
```
Elasticsearch (I am not sure but maybe this is already obtained by the Elasticsearch diagnostics).
```
/_transform/_stats
```

111andre111:
Fixed perPage need for Fleet API calls
Added ability to use parameters in URLs

Closes #512


111andre111:
@pickypg @PhaedrusTheGreek 
If I could get a review on this that would be great. Thanks a lot.

src/main/resources/kibana-rest.yml
```diff
@@ -62,3 +62,43 @@ kibana_user:
   versions:
     ">= 7.6.0": "/internal/security/me"
     ">= 5.0.0 < 7.6.0": "/api/security/v1/me"
+
+kibana_security_exception_list:
+  versions:
+    ">= 7.10.0": "/api/exception_lists/_find?namespace_type=agnostic"
```
peluja1012, line 68: The call being added to this PR will only fetch namespace agnostic exception lists, which in most situations will only include a single "endpoint" exception list. For this reason, could we also add the following call? This will return all of the exception lists associated with the detection rules.

```
"/api/exception_lists/_find?namespace_type=single&per_page=1000"
```

src/main/resources/kibana-rest.yml
```diff
@@ -62,3 +62,43 @@ kibana_user:
...
+
```
peluja1012, line 69: Also, the call below will return the individual exception items associated with rules including logical conditions applied. ([API docs](https://www.elastic.co/guide/en/security/current/exceptions-api-find-exception-items.html)). We would have to first fetch all the exception list ids and then loop over them to fetch the exception list items for each list. Is this possible to do in the support diagnostics framework?
```
GET api/exception_lists/items/_find?list_id=<list-id>
```

src/main/resources/kibana-rest.yml
```diff
@@ -62,3 +62,43 @@ kibana_user:
...
+kibana_security_endpoint_trusted_apps:
+  versions:
+    ">= 7.10.0": "/api/endpoint/trusted_apps"
+
+kibana_security_endpoint_exception_host_isolation:
```
peluja1012, line 74: I would remove the word "exception" from the label. So `kibana_security_endpoint_host_isolation`.

src/main/resources/kibana-rest.yml
```diff
@@ -62,3 +62,43 @@ kibana_user:
...
+  versions:
+    ">= 7.14.0": "/api/exception_lists/items/_find?list_id=endpoint_host_isolation_exceptions&namespace_type=agnostic"
+
+kibana_security_endpoint_exception_event_filters:
```
peluja1012, line 78: Same here. I would remove the word "exception" from the label.

src/main/resources/kibana-rest.yml
```diff
@@ -62,3 +62,43 @@ kibana_user:
...
+  versions:
+    ">= 7.14.0": "/api/exception_lists/items/_find?list_id=endpoint_event_filters&namespace_type=agnostic"
+
+kibana_security_endpoint_exceotion_trusted_apps:
```
peluja1012, line 82: Same here. I would remove the word "exception" from the label.

peluja1012 (commented):
Thanks for adding these additional calls! This is going to greatly help us diagnose more efficiently.

peluja1012:
As part of this PR, could we also add some sorting query parameters to the "find rules" call in order to have the rules sorted to the top?
```
kibana_detection_engine_find:
  versions:
    ">= 7.6.0": "/api/detection_engine/rules/_find?sort_field=enabled&sort_order=asc"
```

peluja1012:
The prepackaged rule call currently doesn't return anything. Could we update it to the following?
```
kibana_detection_engine_signals:
  versions:
    "> 7.10.0": "/api/detection_engine/prepackaged/_status"
```

src/main/resources/kibana-rest.yml
pickypg:
```suggestion
kibana_security_endpoint_trusted_apps:
```

src/main/resources/kibana-rest.yml
pickypg:
```suggestion
kibana_security_endpoint_event_filters:
```

src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -101,7 +103,19 @@ public int runBasicQueries(RestClient client, DiagnosticContext context) throws
...
+        if (m.find()) {
+            res = client.execQuery(String.format("%s&per_page=1", action.getUrl()));
+        }
+        else {
+            if (action.getName().equals("kibana_fleet_agents") || action.getName().equals("kibana_fleet_agent_policies") || action.getName().equals("kibana_fleet_package_policies")) {
+                res = client.execQuery(String.format("%s?perPage=1", action.getUrl()));
+            } else {
+                res = client.execQuery(String.format("%s?per_page=1", action.getUrl()));
+            }
+        }
```
pickypg, line 118: This code is modifying pre-existing bad code, but instead of running a regex against the URL to find if it has a `?` in it, let's explicitly parse it in a function so we can make better sense of it, and not make as many assumptions about future versions of the URL (if they fix the `perPage` difference, we'll want to rename the action for fixed versions instead of updating code):

```java
private void getPageUrl(RestEntry action, int page, int perPage) {
  String actionUrl = action.getUrl();
  String perPageField = "per_page";

  if (
    action.getName().equals("kibana_fleet_agents") ||
    action.getName().equals("kibana_fleet_agent_policies") ||
    action.getName().equals("kibana_fleet_package_policies")
  ) {
    perPageField = "perPage";
  }

  final String querystringPrefix = URI.create(actionUrl).getQuery() != null ? "?" : "&";
  final String params = "page=" + page + "&" + perPageField + "=" + perPage;

  return actionUrl + querystringPrefix + params;
}
```

Then this code could be replaced with:

```java
RestResult res = client.execQuery(getPageUrl(action, 1, 1));
```

Ideally we'd actually use the response for more than just looking up the totals, then asking for the first page _again_, but that's largely what I mean about pre-existing bad code.

src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -130,7 +144,18 @@ public void getAllPages(RestClient client, List<RestEntry> queries, int perPage,
...
+        } else {
+            RestResult res;
+            Pattern p = Pattern.compile( ".*\\?.*" );
+            Matcher m = p.matcher( action.getUrl() );
+            if (m.find()) {
+                return new RestEntry(String.format("%s_%s", action.getName(), page), "", ".json", false, String.format("%s&per_page=%s&page=%s", action.getUrl(), perPage, page), false);
+            } else {
+                return new RestEntry(String.format("%s_%s", action.getName(), page), "", ".json", false, String.format("%s?per_page=%s&page=%s", action.getUrl(), perPage, page), false);
+            }
+        }
```
pickypg, line 158: With the function above, this goes back to being one line:

```java
return new RestEntry(String.format("%s_%s", action.getName(), page), "", ".json", false, getPageUrl(action, page, perPage), false);
```

src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -76,7 +78,7 @@ public int runBasicQueries(RestClient client, DiagnosticContext context) throws
         for (Map.Entry<String, RestEntry> entry : context.elasticRestCalls.entrySet()) {
 
             String actionName = entry.getValue().getName().toString();
-            if (actionName.equals("kibana_alerts") || actionName.equals("kibana_detection_engine_find")) {
+            if (actionName.equals("kibana_alerts") || actionName.equals("kibana_detection_engine_find") || actionName.equals("kibana_security_endpoint_trusted_apps") || actionName.equals("kibana_security_endpoint_exception_host_isolation") || actionName.equals("kibana_security_endpoint_exception_event_filters") || actionName.equals("kibana_security_endpoint_exceotion_trusted_apps") || actionName.equals("kibana_fleet_agents") || actionName.equals("kibana_fleet_agent_policies") || actionName.equals("kibana_fleet_package_policies")) {
```
pickypg, line 81: Perhaps for the future, we should add a parameter to the action itself to know if it's paged so that this can become a boolean check instead of this.

pickypg (changes requested):

src/main/resources/kibana-rest.yml
```diff
@@ -62,3 +62,43 @@ kibana_user:
...
+  versions:
+    ">= 7.10.0": "/api/fleet/agents"
+
+kibana_fleet_agent_policies:
+  versions:
+    ">= 7.10.0": "/api/fleet/agent_policies"
+
+kibana_fleet_package_policies:
+  versions:
+    ">=7.10.0": "/api/fleet/package_policies"
```
pickypg, line 100: We've been sorting these keys alphabetically, so `kibana_f` should be sorted up above.

src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
pickypg: FYI, I created an issue around the idea of different `perPage` parameters for Kibana. https://github.com/elastic/kibana/issues/135490

peluja1012 (changes requested):
Could we please add the following calls? The first one will allow us to get a list of exception lists associated with detection rules in the user's system. The second one will return a list of endpoint exception items. Thank you!

```
// Fetches rule exception lists
"/api/exception_lists/_find?namespace_type=single&per_page=1000"

// Fetches  endpoint exception items
/api/exception_lists/items/_find?list_id=endpoint_list&namespace_type=agnostic&per_page=1000"
```

111andre111:
Done

111andre111 (commented):
Done Requested changes.

src/main/resources/kibana-rest.yml
```diff
@@ -42,19 +59,61 @@ kibana_spaces:
...
+
+kibana_security_endpoint_host_isolation:
+  versions:
+    ">= 7.14.0": "/api/exception_lists/items/_find?list_id=endpoint_host_isolation_exceptions&namespace_type=agnostic"
+
+kibana_security_endpoint_event_filters:
+  versions:
+    ">= 7.14.0": "/api/exception_lists/items/_find?list_id=endpoint_event_filters&namespace_type=agnostic"
+
+kibana_security_endpoint_list:
```
peluja1012, line 78: I would rename this to `kibana_security_endpoint_exception_items` to make the file name more self explanatory.

peluja1012 (approved):
Thanks for the updates, @111andre111!

PhaedrusTheGreek:
@111andre111 , while testing this, one of the new APIs came back 404 on ES 8.3:

```
➜  kibana-api-diagnostics-20220816-183813 git:(main) ✗ cat kibana_detection_engine_signals.json 
Endpoint does not exist.
{"statusCode":404,"error":"Not Found","message":"Not Found"}%     
```

PhaedrusTheGreek (dismissed):
There are a few test failing due to the `getUrl()` changes

```
[ERROR]   TestRunKibanaQueries.testQueriesForKibanaWithHeaders:184 » Diagnostic Could no...
[ERROR]   TestRunKibanaQueries.testQueriesRemovingHeaders:208 » Diagnostic Could not ret...
[ERROR]   TestRunKibanaQueries.testQueriesWithoutHeaders:230 » Diagnostic Could not retr...
```

I haven't been able to work out exactly why, but it's something to do with the perPage key

src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -101,13 +125,13 @@ public int runBasicQueries(RestClient client, DiagnosticContext context) throws
     */
     public void getAllPages(RestClient client, List<RestEntry> queries, int perPage, RestEntry action) throws DiagnosticException {
         // get the values needed to the pagination.
-        RestResult res = client.execQuery(String.format("%s?per_page=1", action.getUrl()));
+        RestResult res = client.execQuery(getPageUrl(action, 1, 100));
```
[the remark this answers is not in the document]
111andre111: Looking at the test codebase lots of these can't even work, because we still test to version 6.5.0 but lots of APIs that we test against only exist in later versions only... These should be fixed in another PR in my opinion and would be too much burden for thi PR.

src/main/resources/kibana-rest.yml
pickypg, line 78:
```suggestion
kibana_security_endpoint_exception_items:
```

Take Pedro's suggestion.

pickypg (approved):

pickypg:
The tests could be corrected to work, but the tests themselves were of no value. The issue was ultimately that the APIs needed to be mocked earlier in the test class.

Given the "quality" of the Kibana portion of the diagnostic, I chose to remove the tests as they were even lower quality.

PR #601, opened 2022-08-12
Include ML trained model info in support diag

droberts195:
Adds /ml/trained_models and /ml/trained_models/_stats to the
support diag from 7.10 onwards.

These provide details of Java models in 7.x but are particularly
important in 8.0 and above as they provide us with information
about the native PyTorch NLP models that a customer might be
requesting support for.


dimitris-athanasiou (approved):
LGTM

pickypg (approved):

Issue #599, opened 2022-07-16
Kibana Diag Output | Paginated Alerts JSON repeats IDs and misses last page


stefnestor:
👋 Howdy! 

I'm experiencing issues on the Kibana diagnostic's [alerts paginated output](https://github.com/elastic/support-diagnostics/blob/main/src/main/resources/kibana-rest.yml#L21). I believe I shouldn't attach my v7.17.5 diagnostic since it has sensitive data, so will outline as best I can.

My cluster has 1493 Kibana(/SIEM) Rules. 

When I run the diagnostic it paginates `kibana_alerts_NUMBER.json` at `per_page:100` appropriately recognizing my `total:1493`. 

However, ( **Problem 1** ) I only ended up with `max(NUMBER)=14` which means I only export 1400 of my expected 1493 Rules. 

Sanity checking further ( **Problem 2** ), I pulled `$ cat kibana_alerts_NUMBER.json | jq '.data[]|.id'` for all exported page `NUMBER` and combined all IDs into a file (had 1400) then performed a `uniq` line command which resulted in 1211 unique Rule IDs. To confirm, I grabbed one of the deduplicated IDs and compared its JSON (on page NUMBER 11 & 14) and they were equivalent.

My guess is the pagination problem may need raised to [Kibana](https://github.com/elastic/kibana/issues?q=is%3Aissue+is%3Aopen+alert+api+pagination) but I wanted to start here since the last page is missed which I believe is directly related to this repositories code. 

TIA!

cactusinthenorth:
@kimholyszko @Heidi-Sager Can you take a look at this one

111andre111:
This should have been fixed in this commit here:
https://github.com/elastic/support-diagnostics/pull/594/commits/db9a072cfc09d9060f16760571147359b929f213
or this PR
https://github.com/elastic/support-diagnostics/pull/594

lucabelluccini:
I _think_ the problem still persists.
In my Kibana I have 686 items in alerts and I was getting 6 files of 100 items, while I would have expected 7.

I've prepared a PR https://github.com/elastic/support-diagnostics/pull/602 but I have no automated tests yet about it.
I did try it against my cloud Kibana.

<img width="1333" alt="image" src="https://user-images.githubusercontent.com/87379/188937845-c038c6cf-7667-4257-9fed-6fa1f414b724.png">

I also use perPage set to 1 to just get the total.

111andre111:
[...]

PR #602, opened 2022-09-07
Fix pagination to be more readable and also include last page

lucabelluccini:
Attempt to fix https://github.com/elastic/support-diagnostics/issues/599

### DIFF
--- a/src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
+++ b/src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
@@ -124,23 +124,17 @@ public class RunKibanaQueries extends BaseQuery {
     * @param  action Kibana API name we are running
     */
     public void getAllPages(RestClient client, List<RestEntry> queries, int perPage, RestEntry action) throws DiagnosticException {
-        // get the values needed to the pagination.
-        RestResult res = client.execQuery(getPageUrl(action, 1, 100));
+        // get the values needed to the pagination (need the total)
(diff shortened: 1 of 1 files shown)


src/main/java/co/elastic/support/diagnostics/commands/RunKibanaQueries.java
```diff
@@ -133,40 +133,39 @@ private String getPageUrl(RestEntry action, int page, int perPage) {
...
-            // Get the first actions page
-            queries.add(getNewEntryPage(perPage, 1, action));
-            // If there is more pages add the new queries
-            if (perPage < total) {
-                int numberPages = (int)Math.ceil(total / perPage);
-                for (int i = 2; i <= numberPages; i++) {
-                    queries.add(getNewEntryPage(perPage, i, action));
-                }
+        String result = res.toString();
+        JsonNode root = JsonYamlUtils.createJsonNodeFromString(result);
```
pickypg, line 160:
```suggestion
        JsonNode root = JsonYamlUtils.createJsonNodeFromString(res.toString());
```

pickypg (approved):
LGTM. Thanks for the tweak. I merged this with main and tweaked it a little myself.

lucabelluccini:
Hello @pickypg 
The modifications are to ensure we have at least one file with 0 alerts right?
Thanks!

pickypg:
> The modifications are to ensure we have at least one file with 0 alerts right?

Yep, that's right.

PR #603, opened 2022-09-07
[Kibana] Fix Broken Endpoints

pickypg:
When Kibana receives a request for the exception lists, it will return a `404` response if there are no exceptions (even though the API call is valid).

This updates the handling to log the error and move on in that scenario.


111andre111:
After a review and some tests, LGTM.

PR #605, opened 2022-09-23
monitoring data extraction from elastic cloud

jalogisch:
Added an example of extracting monitoring data from the elastic cloud. This should reduce the round trips to get the data out of that - or elastic cloud enterprise environments.


pickypg (approved):
LGTM

PR #617, opened 2022-12-20
[HealthAPI] explain has been renamed to verbose in 8.6 and the API is GA in 8.7

andreidan:
Update the Health API paths to reflect the latest changes in Elasticsearch 8.6 and 8.7:
1. Starting with 8.6 the `explain` parameter is named `verbose` (I've omitted specifying it as 
   the default is `true`)
2. We will promote the API to generally available in 8.7 so the path changed to `_health`

### DIFF
--- a/src/main/resources/elastic-rest.yml
+++ b/src/main/resources/elastic-rest.yml
@@ -225,6 +225,8 @@ internal_health:
   versions:
     ">= 8.2.0 < 8.3.0": "/_internal/_health?pretty"
     ">= 8.3.0": "/_internal/_health?pretty&explain"
+    ">= 8.6.0 < 8.7.0": "/_internal/_health?pretty"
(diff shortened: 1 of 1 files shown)


i-a-n (approved):
looks good to me

src/main/resources/elastic-rest.yml
```diff
@@ -225,6 +225,8 @@ internal_health:
   versions:
     ">= 8.2.0 < 8.3.0": "/_internal/_health?pretty"
     ">= 8.3.0": "/_internal/_health?pretty&explain"
+    ">= 8.6.0 < 8.7.0": "/_internal/_health?pretty"
+    ">=8.7.0": "/_health?pretty"
```
idegtiarenko, line 229: With this change, should the upper bound be added to the line 227?

PR #624, opened 2023-04-03
Improve APIs Fleet

lucabelluccini:
TODO


jpcarey:
@lucabelluccini do you have more details on the this PR? We also need confirmation that the added APIs do not contain sensitive data.

pickypg:
Going to close this out pending clarification.

PR #629, opened 2023-05-02
Retrieve top-level data_stream property from indices

VimCommando:
We currently collect `alias`, `settings` and `mappings` but the top-level `data_stream` property isn't in the diagnostic anywhere. This will greatly help automating and visualize data streams in our analyzer.

### Checklist


VimCommando:
Did not mean to open a new PR on this.

PR #646, opened 2023-07-04
[Logstash] Create a new diagnostic template based on the flow metrics

edmocosta:
### Description 
In 8.5-8.7, the Logstash team introduced and then expanded Flow Metrics in the Logstash node stats API, which means they are already available as part of capture from this tool. Flow metrics give us current, recent, and lifetime flow rates for many meaningful metrics, allowing us to compare a pipeline or process's current behavior to itself over time.

Currently, the support diagnostics Analyzer does not yet incorporate Flow Metrics and still requires two-spaced captures to provide meaningful rate-of-flow insights. Additionally, the as-presented rate-of-flow metrics in the analyzer tool are not standardized or cohesive (mixing units or mixing wall-clock vs CPU clock)

We want to create a new diagnostic template, embedded in this support-diagnostics tool that includes process-, pipeline-, and plugin-level Flow Metrics so that it can provide guidance and insights from a single capture.

More details can be found on the meta issue: https://github.com/elastic/logstash/issues/15145

### Changes
These changes should only affect the Logstash remote/local reports

- Added the Apache `freemarker` dependency. It might look like an overkill solution for the given template replacements, but the plan is to also migrate a few existing reports generated by other tools, which will leverage the template system (more info about the project can be found here:  https://github.com/elastic/logstash/issues/15144).
- Added a new resource/config folder named `logstash-diagnostic-templates`, which initially contains a single file `flow_metrics.html.ftlh`. The new `GenerateLogstashDiagnostics` command reads all the template files (.ftlh) inside that folder and writes them to the output artifact naming the output files as `logstash_diagnostic_<template-name>` (removing the `.ftlh` extension)
- The `flow_metrics.html.ftlh` is a single-page application that shows the pipeline and plugin flow metrics.

### Checklist

---
Closes https://github.com/elastic/logstash/issues/15145

### DIFF
--- a/src/main/java/co/elastic/support/diagnostics/commands/GenerateLogstashDiagnostics.java
+++ b/src/main/java/co/elastic/support/diagnostics/commands/GenerateLogstashDiagnostics.java
@@ -49,24 +49,31 @@ public class GenerateLogstashDiagnostics implements Command {
     public void execute(DiagnosticContext context) throws DiagnosticException {
         try {
             final RestClient client = context.resourceCache.getRestClient(Constants.restInputHost);
-            final JsonData jsonData = new JsonData(getNodeInfo(client), getNodeStats(client));
+            final Datasource datasource = new Datasource(new JsonData(getNodeInfo(client), getNodeStats(client)));
             final Configuration configuration = createTemplateConfiguration();
 
             for (String templateFileName : getAllTemplates()) {
-                final String outputFileName = templateFileName.substring(0, templateFileName.lastIndexOf(".ftlh"));
-                final Path outputFilePath = Paths.get(context.tempDir, "logstash_diagnostic_" + outputFileName);
-                try {
-                    final Template template = configuration.getTemplate(templateFileName);
-                    final StringWriter writer = new StringWriter();
-                    template.process(new Datasource(jsonData), writer);
-                    SystemUtils.writeToFile(writer.toString(), outputFilePath.toString());
-                } catch (IOException | TemplateException e) {
-                    logger.error(Constants.CONSOLE, String.format("Could not generate Logstash diagnostic: %s - skipping template.", templateFileName), e);
-                }
+                writeTemplateOutputFile(configuration, templateFileName, datasource, getTemplateOutputFilePath(context.tempDir, templateFileName));
             }
         } catch (Exception e) {
-            logger.info(Constants.CONSOLE, "Unexpected error - bypassing some or all Logstash diagnostics report calls. {}", Constants.CHECK_LOG);
+            logger.info(Constants.CONSOLE, "Unexpected error - bypassing some or all Logstash diagnostics calls. {}", Constants.CHECK_LOG);
         }
+    }
+
+    private void writeTemplateOutputFile(Configuration configuration, String templateFileName, Datasource datasource, Path output) throws DiagnosticException {
+        try {
+            final Template template = configuration.getTemplate(templateFileName);
+            final StringWriter writer = new StringWriter();
+            template.process(datasource, writer);
+            SystemUtils.writeToFile(writer.toString(), output.toString());
+        } catch (IOException | TemplateException e) {
+            logger.error(Constants.CONSOLE, "Could not generate Logstash diagnostic file - skipping template {}.", templateFileName, e);
+        }
+    }
+
+    private Path getTemplateOutputFilePath(String targetDir, String templateFileName) {
+        final String outputFileName = templateFileName.substring(0, templateFileName.lastIndexOf(".ftlh"));
+        return Paths.get(targetDir, "logstash_diagnostic_" + outputFileName);
     }
 
     private Configuration createTemplateConfiguration() throws TemplateModelException {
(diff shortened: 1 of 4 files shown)


src/main/java/co/elastic/support/diagnostics/commands/GenerateLogstashDiagnostics.java
```diff
@@ -0,0 +1,171 @@
...
+            for (String templateFileName : getAllTemplates()) {
+                final String outputFileName = templateFileName.substring(0, templateFileName.lastIndexOf(".ftlh"));
+                final Path outputFilePath = Paths.get(context.tempDir, "logstash_diagnostic_" + outputFileName);
+                try {
+                    final Template template = configuration.getTemplate(templateFileName);
+                    final StringWriter writer = new StringWriter();
+                    template.process(new Datasource(jsonData), writer);
+                    SystemUtils.writeToFile(writer.toString(), outputFilePath.toString());
+                } catch (IOException | TemplateException e) {
+                    logger.error(Constants.CONSOLE, String.format("Could not generate Logstash diagnostic: %s - skipping template.", templateFileName), e);
```
andsel, line 64: Why use `String.format` instead of leveraging logging framework placeholders?

src/main/java/co/elastic/support/diagnostics/commands/GenerateLogstashDiagnostics.java
```diff
@@ -0,0 +1,171 @@
...
+    @Override
+    public void execute(DiagnosticContext context) throws DiagnosticException {
+        try {
+            final RestClient client = context.resourceCache.getRestClient(Constants.restInputHost);
+            final JsonData jsonData = new JsonData(getNodeInfo(client), getNodeStats(client));
+            final Configuration configuration = createTemplateConfiguration();
+
+            for (String templateFileName : getAllTemplates()) {
+                final String outputFileName = templateFileName.substring(0, templateFileName.lastIndexOf(".ftlh"));
+                final Path outputFilePath = Paths.get(context.tempDir, "logstash_diagnostic_" + outputFileName);
```
andsel, line 57: Maybe could be extracted in some method that says "target file from template", such as `reportFileNameFromTemplate`

src/main/java/co/elastic/support/diagnostics/commands/GenerateLogstashDiagnostics.java
```diff
@@ -0,0 +1,171 @@
...
+                try {
+                    final Template template = configuration.getTemplate(templateFileName);
```
andsel, line 59: Maybe could be extracted in method `generateReportUsingCapturedData`

andsel (commented):
Left just a couple of thoughts, that I hope could help in shaping it better.

src/main/java/co/elastic/support/diagnostics/commands/GenerateLogstashDiagnostics.java
edmocosta: It's using the `error(Marker marker, String message, Throwable throwable)` method overload. I can switch it to the parametrized method to leverage the placeholder replacements, but TBH, I find the "exception on the last parameter" not so clear if you don't know L4J's that much.

andsel: This code doesn't need to be so performant, but in general is best to use log4j placeholder, because does a lazy evaluation (if debug level is not match the substitution of the string element, or the "toString", is not executed) while using the `String.format` does an eager evaluation.

src/main/java/co/elastic/support/diagnostics/commands/GenerateLogstashDiagnostics.java
edmocosta: I did a small refactoring on the execute method based on your suggestions! Thanks for reviewing :)

src/main/resources/logstash-diagnostic-templates/flow_metrics.html.ftlh
```diff
@@ -0,0 +1,828 @@
...
+<html lang="en">
+<head>
+    <title>Logstash Diagnostic</title>
+    <meta charset="utf-8"/>
+    <meta content="initial-scale=1, width=device-width" name="viewport"/>
+    <script crossorigin="anonymous"
+            src="https://unpkg.com/react@latest/umd/react.development.js"></script>
+    <script src="https://unpkg.com/react-dom@latest/umd/react-dom.development.js"></script>
+    <script crossorigin="anonymous"
+            src="https://unpkg.com/@mui/material@latest/umd/material-ui.development.js"></script>
```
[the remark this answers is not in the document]
edmocosta: We’re planning a few changes on this template for the next sprints, so we can maybe take the opportunity to also migrate it to EUI.

I will also push a small change on this template file changing the CDN libraries from the `@latest` to the current versions and from `development` to `production.min`.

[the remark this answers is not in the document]
edmocosta: Hey @pickypg, if you've completed your review, do you think we can merge it today? If so, could you please do it for us? I have no permissions on this repository. Thank you!

pickypg (approved):

PR #649, opened 2023-07-12
Update README.md

Gunnerva:
Updated line 444 in readme to add ECE or ECK cluster

### Checklist

- [ ] I have verified that the APIs in this pull request do not return sensitive data

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,7 +441,7 @@ Executing against a remote host, full collection, using an ssh public key file a
 ./diagnostics.sh --host 10.0.0.20 --type remote -u someuser --password --ssl --remoteUser someuser --keyFile "~.ssh/es_rsa" --bypassDiagVerify
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -441,7 +441,7 @@ Executing against a remote host, full collection, using an ssh public key file a
 ./diagnostics.sh --host 10.0.0.20 --type remote -u someuser --password --ssl --remoteUser someuser --keyFile "~.ssh/es_rsa" --bypassDiagVerify
 ```
 
-Executing against a cloud cluster. Note that in this case we use 9243 for the port, disable host name verification and force the type to strictly api calls.
+Executing against a cloud, ECE or ECK cluster. Note that in this case we use 9243 for the port, disable host name verification and force the type to strictly api calls.
```
pickypg, line 444:
```suggestion
Executing against a Cloud, ECE, or ECK cluster. Note that in this case we use 9243 for the port, disable host name verification and force the type to strictly api calls.
```

pickypg (approved):

PR #660, opened 2023-08-03
Drop Support for `tar` files and and `light` mode

pickypg:
This PR does a few things:

- Increments the version from 8.x to 9.0 due to the breaking change of dropping support for tar files
- It drops support for `.tar` and `.tar.gz` files, which also means we drop the `--archiveType` flag
- Adds a new, optional flag `--mode` that allows users to collect a minimal Elasticsearch diagnostic via `--mode light`
  - Kibana and Logstash diagnostics do not honor currently this parameter
- Adds a new `diagnostic_manifest.json` that provides more usable-as-JSON details captured diagnostic
- Removes all uses of `pretty` printing from _all_ collected JSON files
- Corrects the Logstash hot threads file to actually be JSON (by removing `human`) and adding a `.txt` version of the hot threads file
- Makes it so that `logstash-api` can actually be used by removing a `RuntimeException` that served no purpose
- My editor corrected many formatting irregularities, so reviewing without whitespace changes will help
- Uses `./` instead of `/` for the base zip file entries

### Checklist

### DIFF
--- a/src/main/java/co/elastic/support/diagnostics/commands/GenerateDiagnosticManifest.java
+++ b/src/main/java/co/elastic/support/diagnostics/commands/GenerateDiagnosticManifest.java
@@ -22,8 +22,6 @@ import java.util.Map;
 /**
  * Generate a manifest containing the basic runtime info for the diagnostic
  * runtime.
- * Some of the values we get, like the Diagnostic version will be used again
- * downstream.
  */
 public class GenerateDiagnosticManifest implements Command {
    private final Logger logger = LogManager.getLogger(GenerateDiagnosticManifest.class);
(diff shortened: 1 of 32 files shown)


src/main/java/co/elastic/support/diagnostics/commands/GenerateDiagnosticManifest.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Generate a manifest containing the basic runtime info for the diagnostic
+ * runtime.
+ * Some of the values we get, like the Diagnostic version will be used again
+ * downstream.
```
pickypg, line 26:
```suggestion
```

src/main/java/co/elastic/support/diagnostics/commands/GenerateManifest.java
```diff
@@ -16,25 +16,21 @@
...
-import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.TimeZone;
 
+/**
+ * Generate a manifest containing the basic runtime info for the diagnostic
+ * runtime.
+ * Some of the values we get, like the Diagnostic version will be used again
+ * downstream.
```
pickypg, line 26:
```suggestion
```

jpcarey (approved):

PR #667, opened 2023-09-08
Add profiling status API to ES diagnostics

danielmitterdorfer:
With this commit we add a new API call to gather the status of Universal Profiling in the diagnostic dump. The API is always available in the default distribution even if the Universal Profiling plugin is disabled, so it is always safe to call.

### Checklist


pickypg (approved):
LGTM

PR #682, opened 2024-02-19
Remove leading `./` from zip entries

DaveCTurner:
The leading `.` on zip entry names is ignored by (at least) the Mac OS
`unzip` utility which means that every file appears to have an absolute
path, so unpacking a diagnostics bundle emits a warning for every entry:

    warning:  stripped absolute path spec from ...

There's no need to prefix every entry name with `./`, so this commit
removes the prefix to avoid these warnings.

### DIFF
--- a/src/main/java/co/elastic/support/util/ArchiveUtils.java
+++ b/src/main/java/co/elastic/support/util/ArchiveUtils.java
@@ -30,7 +30,7 @@ public class ArchiveUtils {
       try (
             FileOutputStream fout = new FileOutputStream(filename);
             ZipArchiveOutputStream taos = new ZipArchiveOutputStream(fout)) {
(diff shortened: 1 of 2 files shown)


andymjames (commented):
Hey X-School friend! I left for a while, but I'm back!

This looks pretty straight-forward, but since I never got too familiar with this particular project, I am going to get some extra eyes on here and guidance on how/if we test a PR like this.

I also feel like I need to make sure that nobody has some strong objection to the change (but I think you made a pretty good case for why this change would be good idea).

DaveCTurner:
Hey @andymjames it's been a while, welcome back! No rush on this, it's just got to me enough times to chase down the problem :)

pickypg (approved):

PR #721, opened 2024-08-20
Pagination refactor + Space awareness for Kibana

lucabelluccini:
Attempt to solve https://github.com/elastic/support-diagnostics/issues/578

### Checklist

- Test with older versions of Kibana
- Review the list of APIs and make them paginated, space aware (or both)

### Notes/Obs

The KibanaUrl code could be integrated in the RestEntry - if we agree on that we can change it.

---

Hold prior merging.


src/main/java/co/elastic/support/diagnostics/commands/KibanaUrl.java
```diff
@@ -0,0 +1,60 @@
...
+        this.url = inputUrl;
+    }
+
+    public static KibanaUrl parse(String inputUrl) {
+        boolean isPaginated = false;
+        String paginationFieldName = null;
+        boolean isSpaceAware = false;
+
+        // Single regex pattern to match the end of the string for both cases
+        String pattern = "(.*?)(#spaceaware)?(#paginate\\[(.*?)\\])?$";
```
pickypg, line 23: Instead of doing this, we should just add a companion field.

src/main/resources/kibana-rest.yml
```diff
@@ -20,7 +24,7 @@ kibana_actions:
 
 kibana_alerts:
   versions:
```
pickypg, line 26: Something like

```suggestion
kibana_alerts:
  paginate: true
  spaceaware: true
  versions:
```

src/main/resources/kibana-rest.yml
```diff
@@ -20,7 +24,7 @@ kibana_actions:
...
-    ">= 7.9.0": "/api/alerts/_find"
+    ">= 7.9.0": "/api/alerts/_find#spaceaware#paginate[per_page]"
```
pickypg, line 27: Or perhaps to future-proof it:

```suggestion
kibana_alerts:
  versions:
    ">= 7.9.0": "/api/alerts/_find#spaceaware#paginate[per_page]"
      paginate: per_page
      spaceaware: true
```

lucabelluccini: I can go for this one

lucabelluccini:
Thank you @pickypg for the review - I'll work on https://github.com/elastic/support-diagnostics/pull/721#discussion_r1724017406

src/main/resources/kibana-rest.yml
lucabelluccini: Actually, this is not valid YAML - I have to go with:
```
kibana_alerts:
  versions:
    ">= 7.9.0":
      api: "/api/alerts/_find#spaceaware#paginate[per_page]"
      paginate: per_page
      spaceaware: true
```

lucabelluccini: I've submitted a change which allows both the old format and the new format.

pickypg: Thanks. I'll review this today.

pickypg (approved):
I am going to merge this and make a few tweaks to push.

Issue #676, opened 2023-11-06
Add Kibana Reporting diagnostics


tsullivan:
Reporting has a self-service diagnostic tool in Stack Management > Reporting. 

From [the docs](https://www.elastic.co/guide/en/kibana/current/reporting-troubleshooting.html#:~:text=Reporting%20diagnosticsedit&text=When%20Kibana%20is%20running%2C%20navigate,up%20with%20any%20relevant%20recommendations.):
> [In the] the Report Listing page, and click Run reporting diagnostics. This will open up a diagnostic tool that checks various parts of the Kibana deployment and come up with any relevant recommendations

We should have this Reporting Diagnostic get pulled as part of the general Kibana diagnostic. It would help speed up discovery and timelines for the Support team.

tsullivan:
Replaces: https://github.com/elastic/kibana/issues/151033

tsullivan:
The diagnostic tool is under **Stack Management > Alerts and Insights > Reporting > Run reporting diagnostics.**

<img width="1701" alt="image" src="https://github.com/elastic/support-diagnostics/assets/908371/4d6e0853-dd97-4575-8175-48bb9ae3ffe2">

The client request to the Reporting Diagnostic API ("Check browser" functionality):
```
curl 'https://localhost:15601/api/reporting/diagnose/browser' \
  -X 'POST' \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-US,en;q=0.9' \
  -H 'Connection: keep-alive' \
  -H 'Content-Length: 0' \
  -H 'Content-Type: application/json' \
[...]

PR #746, opened 2024-09-21
[Kibana] Add Reporting `/internal/reporting/diagnose/browser`

pickypg:
Should the requirements for communicating with this API ever change (e.g, we have to specify a header), then we may need to modify it in the future, but for now it works as-is.

Closes https://github.com/elastic/support-diagnostics/issues/676

### Checklist

### DIFF
--- a/src/main/resources/kibana-rest.yml
+++ b/src/main/resources/kibana-rest.yml
@@ -106,6 +106,10 @@ kibana_fleet_settings:
   versions:
     ">= 7.14.0": "/api/fleet/settings"
 
+kibana_reporting_diagnose_browser:
+  versions:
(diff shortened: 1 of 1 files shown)


pickypg:
@tsullivan If you could verify the PR statement and check the box, that would be helpful and then we can get this out sometime next week. :)

tsullivan:
@pickypg are log entries that may contain file paths considered sensitive data?

pickypg:
No, it's more about logging user credentials or actual data.

tsullivan:
Thanks, I verify that the API does not return sensitive data. However, I can't check the checkbox as I do not have write privilege to update the description text.

Issue #657, opened 2023-07-26
Set new header for diag requests to Kibana on serverless


jloleysens:
For serverless we plan on making most of Kibana's HTTP APIs "internal". This means they will be soft blocked: requiring setting a [new special header](https://github.com/elastic/kibana/blob/c0cb6133eae295dbf05151e2dd4ebafc391e9f64/packages/core/http/core-http-common/src/constants.ts#L13) otherwise we return 400.

**Note**: this will only be enforced for serverless, onprem will not be restricted in this way. 
**Additional note**: setting the header will not cause issues for our 

"Internal" in this case refers only to HTTP API users that want to integrate with Kibana. "Public" endpoints will be versioned, documented and be very long lived. "Internal" endpoints , especially those used by our tooling for diagnostics are not necessarily expected to change often -- we just want to prevent end users from knowing about/integrating with them given the much larger commitment that goes with making them public.

My suggestion is rather than making these HTTP APIs "public" in the sense described above, we give this tool access to "internal" HTTP APIs so that it can be used against serverless Kibana's in the same way.

Let me know what you think!

CC @xcrzx

jloleysens:
Related https://github.com/elastic/kibana/pull/162258

111andre111:
It seems we will mark for serverless APIs as internal as well making them available using the same header:
https://github.com/elastic/elasticsearch/pull/93607

banderror:
> Note: this will only be enforced for serverless, Cloud/onprem will not be restricted in this way.

As far as I understand this is not true. All endpoints with `access: 'internal'` will return 400 if the version header is not passed, in all types of environments: Serverless and ESS.

This means that the support-diagnostics tool will not work with a large number of Kibana endpoints.

pickypg:
> All endpoints with `access: 'internal'` will return 400 if the version header is not passed

In the highlighted code, it's highlighting the `x-elastic-internal-origin` header. Was the intended header reference to meant to be `elastic-api-version`?

banderror:
@pickypg The tool should send both headers in HTTP requests to Kibana API endpoints:

- The `x-elastic-internal-origin` can be passed to all endpoints with the same constant value, e.g. `x-elastic-internal-origin: support-diagnostics`
- The value for the `elastic-api-version` header should be (optionally) specified separately for each endpoint + version in the [YAML config](https://github.com/elastic/support-diagnostics/blob/main/src/main/resources/kibana-rest.yml) file, for example:

```yaml
[...]

PR #756, opened 2024-10-08
Add per request functionality to add HTTP request headers

111andre111:
### Feature statement

Adds functionality to configure headers on a per request statement
Syntax example:
```
<requestname>:
  versions:
    ">= <stackversion>":
      url: "<api url"
      extra-headers:
        <header1>: "<header1value"
        <header2>: "<header2value" 
```

### Checklist

- [x] I have verified that the existing global header functionality in `config/diags.yml` from #519 still works
e.g.
```
extra-headers:
  elastic-api-version: "2023-10-31"
```
- [x] I have verified that the local headers in `{kibana|elastic|logstash}-rest.yaml` overwrite the global header functionality

Fixes elastic/support-diagnostics#657

### DIFF
--- a/src/main/java/co/elastic/support/rest/RestEntryConfig.java
+++ b/src/main/java/co/elastic/support/rest/RestEntryConfig.java
@@ -74,20 +74,31 @@ public class RestEntryConfig {
             if (semver.satisfies(urlVersion.getKey())) {
                 if (urlVersion.getValue() instanceof String) {
                     return new RestEntry(name, subdir, extension, retry, (String) urlVersion.getValue(), showErrors);
-                    // We allow it to be String,String or String,Map(url,paginate,spaceaware)
+
                 } else if (urlVersion.getValue() instanceof Map) {
                     Map<String, Object> info = (Map<String, Object>) urlVersion.getValue();
-
                     String url = (String) ObjectUtils.defaultIfNull(info.get("url"), null);
-
                     if (url == null) {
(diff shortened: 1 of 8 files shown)


src/main/java/co/elastic/support/rest/RestEntryConfig.java
```diff
@@ -74,24 +74,35 @@ private RestEntry buildRestEntryForVersion(String name, Map<String, Object> entr
...
-
                     String pageableFieldName = (String) ObjectUtils.defaultIfNull(info.get("paginate"), null);
                     boolean spaceAware = (boolean) ObjectUtils.defaultIfNull(info.get("spaceaware"), false);
 
-                    return new RestEntry(name, subdir, extension, retry, url, showErrors, pageableFieldName, spaceAware);
+                    // Construct the RestEntry object first
+                    RestEntry restEntry = new RestEntry(name, subdir, extension, retry, url, showErrors, pageableFieldName, spaceAware);
+
+                    // Logic for handling extra-headers, if they exist, after RestEntry is created
+                    if (info.containsKey("extra-headers")) {
```
pickypg, line 91: Is there a downside to just extracting the headers during creation, like we do with `spaceAware` for example?

pickypg: @111andre111 Is there a reason for this or just how you implemented it at the time (either way is fine, I'm just trying to understand)?

PR #772, opened 2024-12-02
[Logstash] Add Health Report API output

lucabelluccini:
Following https://github.com/elastic/logstash/pull/16520 and consequent https://www.elastic.co/guide/en/logstash/current/health-report-pipeline-flow-worker-utilization.html, I'm adding this API.

FYI @robbavey / @jsvd - I'm adding this API to the Support Diagnostic tool.
Is there any other way to leverage it's output? Should we add something to the Flow visualizer or our Logstash Visualizer?

### Checklist

- [X] I have verified that the APIs in this pull request do not return sensitive data

### DIFF
--- a/src/main/resources/logstash-rest.yml
+++ b/src/main/resources/logstash-rest.yml
@@ -38,3 +38,7 @@ logstash_nodes_hot_threads_human:
 logstash_plugins:
   versions:
     "> 0.0.0": "/_node/plugins"
+
+logstash_health_report:
+  versions:
+    "> 8.16.0": "/_health_report"


src/main/resources/logstash-rest.yml
```diff
@@ -38,3 +38,7 @@ logstash_nodes_hot_threads_human:
 logstash_plugins:
   versions:
     "> 0.0.0": "/_node/plugins"
+
+logstash_health_report:
```
pickypg, line 42: Can we sort this so it's in a predictable spot?

pickypg (changes requested):

src/main/resources/logstash-rest.yml
lucabelluccini: Should I sort the entire list of APIs?

pickypg: Ah, no. I hadn't realized Logstash had not been sorted properly. I'll merge this, then do it all.

pickypg (approved):

PR #793, opened 2025-02-24
logstash-flow-metrics added for logstash-api diag

### LINKED ISSUE #792: [Logstash] Add flow metrics to type logstash-api diagnostics as well
Hi, is there a reason that currently only `logstash` and `logstash-remote` support the flow metrics diagnostics generation for Logstash diagnostics?

I made some tests, I could not find a reason why this should not work with api diagnostics as well.

This is referring to:
https://github.com/elastic/logstash/issues/15145
and 
https://github.com/elastic/support-diagnostics/pull/646

I have opened a PR against this issue, that should actually implement it for logstash api diags as well.

alstolten:
### Checklist

-> no change in API requests, just added the same set to `logstash-api` type diagnostics.

fixes #792


lucabelluccini:
Thanks! That's great

pickypg (approved):
LGTM

Issue #795, opened 2025-03-04
Support pre-release versions


VimCommando:
When running the diagnostic tool against a `9.0-rc1` or `9.0-SNAPSHOT` build it fails to collect any data. I've seen this in the past with 8.x snapshot builds as well.

This was collected from an ESS cluster:

```jsonc
// diagnostic_manifest.json
{
  "mode" : "full",
  "product" : "elasticsearch",
  "flags" : "BaseInputs{outputDir='/tmp/diagnostics-13164653263138019883'},ElasticRestClientInputs: {host='8fba3854e506471fa5a22d0deb7980a5.us-west2.gcp.elastic-cloud.com', port=9243, proxyHost='', proxyPort=8080, user='', isPassword=false, pkiKeystore='', isPkiPass=false, pkiKeystorePass='', proxyUser='', isProxyPass=false, isSsl=true, skipVerification=false, scheme='https'},DiagnosticInputs: {, diagType='api', mode='full', remoteUser='null', keyfile='', isKeyFilePass=false', trustRemote=false', knownHostsFile='', sudo=true', remotePort=22, runner='ess'}",
  "diagnostic" : null,
  "type" : "elasticsearch_diagnostic",
  "runner" : "ess",
  "version" : "9.0.0-rc1",
  "timestamp" : "2025-03-04T18:59:50.537Z"
}
```

```jsonc
// manifest.json
{
  "diagnosticInputs" : "BaseInputs{outputDir='/tmp/diagnostics-13164653263138019883'},ElasticRestClientInputs: {host='8fba3854e506471fa5a22d0deb7980a5.us-west2.gcp.elastic-cloud.com', port=9243, proxyHost='', proxyPort=8080, user='', isPassword=false, pkiKeystore='', isPkiPass=false, pkiKeystorePass='', proxyUser='', isProxyPass=false, isSsl=true, skipVerification=false, scheme='https'},DiagnosticInputs: {, diagType='api', mode='full', remoteUser='null', keyfile='', isKeyFilePass=false', trustRemote=false', knownHostsFile='', sudo=true', remotePort=22, runner='ess'}",
  "diagVersion" : null,
  "Product Version" : {
    "major" : 9,
    "minor" : 0,
    "patch" : 0,
    "preRelease" : [ "rc1" ],
    "build" : [ ],
    "version" : "9.0.0-rc1",
    "stable" : false
  },
  "runner" : "ess",
[...]

PR #796, opened 2025-03-07
Workaround for the semver issue with pre-release versions

dimuon:
We came across an issue with failing tests in ECE when the tool is used against pre-release ECE stack versions (e.g. "9.0.0-beta1"). 
We tracked it down to [this semver defect](https://github.com/semver4j/semver4j/issues/307). 

The PR workarounds the problem by removing the pre-release part from version. 
It should not affect the usual workflow though but it can help us to reduce the number of failing tests in CI pipelines until the root cause get fixed.

Closes https://github.com/elastic/support-diagnostics/issues/795

### Checklist

- [] I have verified that the APIs in this pull request do not return sensitive data

Rel: https://elasticco.atlassian.net/browse/CP-10456

### DIFF
--- a/src/main/java/co/elastic/support/diagnostics/commands/CheckElasticsearchVersion.java
+++ b/src/main/java/co/elastic/support/diagnostics/commands/CheckElasticsearchVersion.java
@@ -89,6 +89,8 @@ public class CheckElasticsearchVersion implements Command {
         String result = res.toString();
         JsonNode root = JsonYamlUtils.createJsonNodeFromString(result);
         String version = root.path("version").path("number").asText();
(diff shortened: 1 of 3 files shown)


pickypg (approved):
Thanks for the fix!

pickypg:
@dimuon I've done a release (`9.2.3`) containing this fix.

https://github.com/elastic/support-diagnostics/releases/tag/v9.2.3
https://repo1.maven.org/maven2/co/elastic/support/diagnostics/ (currently not updated in the Maven caches yet)

dimuon:
@pickypg , thanks!

PR #832, opened 2025-05-05
Add tests and CI support for automatic testing for Python scripts

valeriy42:
This PR adds unit tests for scripts and automatic PR testing with GitHub workflow. This will ensure that automatic dependency updates will not break the behaviour.


valeriy42:
@pickypg can you please give it a review?

PR #989, opened 2026-04-16
Update Logstash report to include batch's byte_size p100 and lifetime

### LINKED ISSUE #19008: update diag tool to take this information into the html report it generates

andsel:
Updates the report generated when taking Logstash diags to include the max percentile just for batch's byte_size. Furthermore, adds the lifetime window for all batch relates metrics (byte_size and event_count).

### Checklist

## What the PR does?
Updates the HTML template file used to generate the report bundled with Logstash diagnostic to add the lifetime window column for all percentiles of batch structure metric.
In the same table add a new percentile, p100 or max which is available only for batch's byte_size metric.

Following a set of snapshots of the updated section:

Screenshot 2026-04-16 at 10 52 04

- Closes https://github.com/elastic/logstash/issues/19008

### DIFF
--- a/src/main/resources/logstash-diagnostic-templates/flow_metrics.html.ftlh
+++ b/src/main/resources/logstash-diagnostic-templates/flow_metrics.html.ftlh
@@ -159,6 +159,30 @@
         }
     }
 
+    const _batchByteSizeAverageRolling = (pipelineName) => {
+        try {
+            const a = data.stats.pipelines[pipelineName].batch.byte_size.average;
+            if (!a) return "N/A";
+            for (const k of ["last_1_minute", "last_5_minutes", "last_15_minutes"]) {
+                const v = a[k];
+                if (v !== null && v !== undefined && v !== "") return v;
+            }
+        } catch (e) {}
+        return "N/A";
+    }
+
+    const _batchEventCountAverageRolling = (pipelineName) => {
+        try {
+            const a = data.stats.pipelines[pipelineName].batch.event_count.average;
+            if (!a) return "N/A";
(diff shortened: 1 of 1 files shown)


lucabelluccini (approved):
😍

src/main/resources/logstash-diagnostic-templates/flow_metrics.html.ftlh
```diff
@@ -159,6 +159,30 @@
         }
     }
 
+    const _batchByteSizeAverageRolling = (pipelineName) => {
+        try {
+            const a = data.stats.pipelines[pipelineName].batch.byte_size.average;
```
pickypg: These functions are identical except the second-to-last field being accessed here. You could make that a parameter, or you could accept a lambda callback function.

Like

```js
   const _batchAverageRolling = (pipelineName, fieldName) => {
        try {
            const a = data.stats.pipelines[pipelineName].batch[fieldName].average;
```

then just call it with `_batchAverageRolling(selectedPipeline.name, 'byte_size')` instead of how it's done currently. Then you can use this for other fields, but also avoid the duplicated function (if there's ever a bug in one, then there's a bug in the other, and the person fixing it has to remember to fix both as-is).

pickypg (approved):
This looks good, but I think you can / should merge these functions to avoid duplication.

src/main/resources/logstash-diagnostic-templates/flow_metrics.html.ftlh
andsel: Thanks @pickypg for your thoughts, I've integrated your suggestion and it's ready for review 🙏

pickypg (approved):

PR #1033, opened 2026-07-07
Add Kibana Streams API (/api/streams, /api/streams/_status) to diagnostics

lucabelluccini:
## Summary
- Adds `kibana_streams` (`GET /api/streams`) to `kibana-rest.yml` so the diagnostic collects Kibana Streams configuration.
- Adds `streams_status` (`GET /_streams/status`) to `elastic-rest.yml` so the diagnostic collects stream enablement status directly from Elasticsearch. Kibana's own `GET /api/streams/_status` is a thin proxy over this Elasticsearch endpoint and was rejected by at least one Kibana build ("exists but is not available with the current configuration"), so we query Elasticsearch directly instead.
- Both are version-gated to `>= 8.19.0 < 9.0.0 || >= 9.1.0`, matching where these APIs actually exist (confirmed against the 8.19.0/9.0.0/9.1.0 release tags for both Kibana and Elasticsearch — the APIs are present in the 8.19.0 and 9.1.0 trains, but not in 9.0.x).
- Neither endpoint supports pagination.

## Test plan
- [x] `./gradlew compileJava` succeeds
- [x] `./gradlew test --tests "co.elastic.support.rest.*"` passes (existing YAML-loading tests still pass with the new entries)
- [x] Built the dist zip and ran `kibana-api` diagnostics against a local Kibana 9.5.0 instance — `kibana_streams.json` collected real stream data successfully
- [x] Manually run diagnostics against a Kibana/Elasticsearch 9.1.0+ (or 8.19.0+) instance and confirm `streams_status.json` is produced
- [x] Confirm the entries are correctly skipped on 9.0.x


pickypg:
@lucabelluccini I'll wait until you can confirm those last two, then I can merge this.

lucabelluccini:
Hello @flash1293 / @3kt  - May I ask if those APIs make sense and if we should gate them on a 9.2 or even 9.4 version instead?
Can those APIs leak PII/data?

lucabelluccini:
Hello @pickypg - I need a review by Joe/Alexis - Let's hold please

Tests run against 9.5.0-SNAPSHOT
```
Elasticsearch (streams_status.json, from GET /_streams/status):
{"logs":{"enabled":false},"logs.ecs":{"enabled":true},"logs.otel":{"enabled":true}}

Kibana (kibana_streams.json, from GET /api/streams):
{"streams":[{"type":"wired","name":"logs.ecs",...},{"type":"wired","name":"logs.otel",...},{"type":"classic","name":"logs-elastic_agent-default",...}, ...]}
```

flash1293 (approved):
LGTM, /api/streams is not super light, but since the number of streams is bounded by what Elasticsearch can handle (a couple thousand at max), it shouldn't cause crazy memory pressure.

pickypg (approved):

PR #1042, opened 2026-07-13
[Buildkite] Require Hash with comment

pickypg:
This limits the usage of the fork-CI execution to running the explicitly defined hash that must exist before the comment.


pickypg:
buildkite test this fake

pickypg:
(Woops that only tested `main`)

